PR #75, opened 2015-11-30
Docs: Fix a few style bugs

paulathevalley:
Resolves https://github.com/FormidableLabs/formidable-landers/issues/27 

- Fix sticky footer in Safari
- Remove duplicated `<Style>` tag 
- Find source of the horizontal scrollbars: the triple ticks notation wraps everything in `<pre>` tag, as per github flavored markdown rules, but this tag is the cause. For instance, if this tag was changed to `<div>` there would be no horizontal scrollbars, but I did not find a conventional way to add a class name to a `<div>` in markdown notation. 

/cc @david-davidson @boygirl


david-davidson:
@paulathevalley Yeah buddy! :ship:

PR #76, opened 2015-12-01
[Doc site] WIP: incorporate component docs directly in site!

david-davidson:
This adds the infrastructure to support including specific component docs in the overall Victory site, `require`d in directly from those components.

It's waiting on three things:
* Version bumps for `victory-axis`, `victory-chart`, and `victory-pie` (the fix for importing their docs without automatically rendering them).
* Some styling work on the specific component pages--they're looking a little compressed. @paulathevalley, what if we gave the sidebar a little less room on the left?
* Minor but odd: the first few ecology docs for `victory-bar` are rendering `TypeError: Cannot read property 'length' of undefined` in the live preview. Don't think it's part of this PR, but we'll need to solve it before rolling this out.

I've commented out the routes / sidebar items for the components that don't have the version bump yet, so you should be able to run `npm run dev-docs` and click around to your heart's content. (Requires wiping old victory packages, `npm install`ing again.)

/cc @paulathevalley @boygirl -- since I'll be out of the office for the rest of the week, feel free to take over!


paulathevalley:
Flex was setting a max width (no clue how that width got determined), but I was able to adjust it so that the docs can go wider: 

screen shot 2015-12-01 at 2 16 19 pm

The rest of the victory component repos will need to remove the `.Container` wrapper class to also get these changes (will do that now).

boygirl:
: ( 
I just added all of the docs to `.npmignore`. so this will break when those changes are released. We could talk about keeping just the `ecology.md` and props json blobs in the packages, and building them all here though.

boygirl:
@david-davidson after a discussion with @coopy I'm going to leave `docs` in the npm package, and just npm ignore `docs/build`

boygirl:
@paulathevalley all of the dependencies have new versions, so this shouldn't be blocked anymore. I also cant reproduce the issue on victory-bar

docs/components/component-docs.jsx
```diff
@@ -0,0 +1,56 @@
...
+import Sidebar from "./sidebar";
+
+import BaseDocs from "./docs";
+
+import theme from "./theme";
+
+const docs = {
+  // "victory-axis": require("victory-axis/docs/docs"),
+  // "victory-chart": require("victory-chart/docs/docs"),
+  // "victory-pie": require("victory-pie/docs/docs"),
```
coopy: When will we include these? Are there issue #'s?

boygirl:
@paulathevalley @coopy 
I updated all the deps, and moved to using formidable-landers assets, but I cant `npm run dev-docs` to run on this branch. I get the following error:
```
Uncaught Error: only one instance of babel/polyfill is allowed
```

Any takers?

PR #79, opened 2015-12-01
use breakConfig option in .babelrc file to avoid inheriting babelrc from parent project

dandelany:
This addresses #78 by adding a .babelrc file with `{"breakConfig": true}` to Victory. Usually, babel looks for a .babelrc file in its current directory and parent directories, and *merges* them all together to get a combined Babel configuration. However, we don't want this behavior in this case, because when the `babel` postinstall step runs, it ends up using the .babelrc from the parent (installing) project if it exists, which may contain options we don't want.

`breakConfig` is an undocumented babel 5 option which causes babel to *only* use this babelrc file and *not* inherit/merge configuration files from parent directories. Tested this and it seems to fix my issue noted in #78. Don't think it should have any unintended consequences, though I'm not sure - you guys know your build process better than I do obviously :) LMK if you have any questions. Thanks,

-d


exogen:
@dandelany Awesome! Was looking for exactly an option like this, but the documentation is lacking. I won't have time to test this out until later, but it looks good :+1:

baer:
Will this still work in Babel@6? Upgrading seems eminent.

exogen:
@baer It's not needed in Babel 6 (and our `.babelrc` will have to change significantly anyway, Babel 6 errors out on all the old options) – this is just to get us by in the meantime so Babel 5 isn't broken.

baer:
:+1:

exogen:
@dandelany: Works great! Can you merge/rebase onto master, which will require adding the `stage` and `nonStandard` options too? That way it matches our other repos and if we eventually use something like the `@Radium` decorator in `app.jsx`, it'll be ready to handle it.

dandelany:
Apologies for the delay - just merged master into my branch and updated the babelrc with the new options. Tested a fresh install on my project with Babel 6, and it seems to work now. Should be good to merge.

PR #103, opened 2015-12-16
Remove playgrounds from /docs index, and add link to Victory GitHub

### LINKED ISSUE #70: Docs site: better example of importing Victory components
Right now, when we describe how to use Victory, we rely on commented-out code in the ecology playground:
```
// import { VictoryPie } from "victory"
<VictoryPie/>
```
We should put a non-commented-out `import` statement in its own code block above ecology, or find another way to demonstrate usage without relying on dead code.

/cc @boygirl--if you feel like tinkering with this one, go for it!

david-davidson:
The playgrounds currently up at http://victory.formidable.com/docs/ are kind of boring. They're left over from when the awesome, non-boring component-specific docs weren't actually part of the site. Also, @alexlande wisely points out that we lack a direct link back to the GitHub repo.

So: this removes the old playgrounds and "Getting started" steps, in favor of a few quick links. I figure people would rather get straight to the component docs anyway.

/cc @coopy @boygirl @alexlande @eastridge 

It looks like this:
victory-docs

### DIFF
--- a/docs/components/docs.jsx
+++ b/docs/components/docs.jsx
@@ -6,12 +6,8 @@ import React from "react";
 import ReactDOM from "react-dom";
 
 import { components } from "../config";
-import { VictoryTheme, Header, Footer } from "formidable-landers";
-import * as Victory from "../../src/index";
-const { VictoryChart, VictoryLine, VictoryPie } = Victory;
-const V = Victory;
-
 import Sidebar from "./sidebar";
+import { VictoryTheme, Header, Footer } from "formidable-landers";
 
 @Radium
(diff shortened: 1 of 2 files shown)


eastridge:
I don't think this leaves anything meaningful on the page. I would suggest instead adding a link to github and to the list of component docs on the front page.

david-davidson:
@eastridge Yeah, it wouldn't leave much--I'll just add the link.

david-davidson:
Closing this in favor of a bulk PR with other last-minute tweaks.

PR #133, opened 2015-12-28
Exclude dist

### LINKED ISSUE #131: Exclude dist from the repository, but keep it in the npm
As per https://github.com/FormidableLabs/radium/issues/471#issuecomment-167003690 I think we should do that in all our projects. There is no point of keeping track of `dist` on GH.

### Solution
As described in the comment above:
- add `dist` to `.gitignore`
- create empty `.npmignore` which overrides `.gitignore` so the `dist` will be included in the npm package and can be used via https://npmcdn.com

More in the [npm docs](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package).

knowbody:
fixes #131


ryan-roemer:
@knowbody -- `postinstall` now needs to build _both_ `lib` and `dist`. Please update in archetype first.

knowbody:
oops forgot about it, good catch, thanks @ryan-roemer

knowbody:
cc @ryan-roemer @boygirl I resolved the conflicts, so it can be merged now

PR #149, opened 2016-01-04
Docs: Use default copy in header and set maxWidth to props table

paulathevalley:
Based on https://github.com/FormidableLabs/formidable-landers/pull/45

- Use default copy for header
- Set an explicit `backgroundColor` prop for the header and footer
- Set `maxWidth` for the props type column: 
screen shot 2016-01-04 at 9 47 43 am

/cc @coopy @david-davidson

### DIFF
--- a/docs/components/app.jsx
+++ b/docs/components/app.jsx
@@ -29,7 +29,7 @@ class App extends React.Component {
   render() {
     return (
       <div style={{display: "flex", minHeight: "100vh", flexDirection: "column"}}>
-        <Header backgroundColor="#ebe3db">
+        <Header backgroundColor={VictorySettings.palestSand}>
           Looking for custom dashboards or data visualization consulting? Let&rsquo;s talk.
         </Header>
         <main className="Container" style={this.getMainStyles()}>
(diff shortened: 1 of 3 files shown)


boygirl:
@paulathevalley LGTM

paulathevalley:
@boygirl :+1: I sneakily made two more changes. As requested by Tolis, added a TM to the first use of the word, "Victory," and added a footnote in the footer (the alternative would be to add a TM to every instance of VICTORY throughout the site). 

screen shot 2016-01-04 at 3 41 39 pm

screen shot 2016-01-04 at 3 41 45 pm

paulathevalley:
- Update the copy in the header to be specific to victory

screen shot 2016-01-04 at 4 33 11 pm

docs/components/app.jsx
```diff
@@ -29,9 +29,9 @@ class App extends React.Component {
   render() {
     return (
       <div style={{display: "flex", minHeight: "100vh", flexDirection: "column"}}>
-        <Header
-          text={"Interested in using Victory on your next project? Let’s talk."}
-        />
+        <Header backgroundColor="#ebe3db">
```
coopy: Please use `VictorySettings.palestSand` instead of the literal value. That goes for all `backgroundColor` props below as well.

coopy:
Great! :+1:

david-davidson:
Yeah, looks :ok_hand:!

PR #158, opened 2016-01-13
Update header; Add Roadmap link

coopy:
- Update header text in three places
  - Override `display:flex` since that messes up the underline.
  - Now the string lives in docs/config, for easier updates in the future.
- Add Roadmap link
- Update dev documentation for docs


david-davidson:
@coopy Aside from the single quotes, LGTM! :sparkles:

PR #164, opened 2016-01-13
Use AND instead of OR in postinstall

austinpray:
The pipes here looked odd.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   },
   "homepage": "https://github.com/formidablelabs/victory",
   "scripts": {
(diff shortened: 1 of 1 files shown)


exogen:
@austinpray It's intentional, here's how it breaks down:

`cd lib`: If this works, then `lib` exists and is a directory; so do nothing. If it fails, then `||` will cause us to run the next command...

`builder run npm:postinstall`: This builds `lib`, which we only need to do if the `lib` directory does *not* exist, i.e. the previous command failed.

This lets us support `git` install URLs, since `lib` is built into the npm package but not checked into source control. (As noted in the warning however, trying to build stuff in `postinstall` can still cause issues in some scenarios, so we've had a lot of discussion around how to improve this.)

austinpray:
Ah makes sense.

exogen:
@austinpray Thanks for looking into it either way! :)

PR #195, opened 2016-02-04
add radium grid to the navbar, with custom set sizes and breakpoints

mdaxtman:
@coopy @paulathevalley @tptee 

Simple update to the victory docs. Add grids and cells to the navbar


mdaxtman:
unnecessary changes as per Per's instructions, closed.

PR #219, opened 2016-03-03
Move url- and style-loader deps to archetype

coopy:
Dependent on https://github.com/FormidableLabs/builder-victory-component/pull/36


coopy:
This actually just removes the `url-loader` and `style-loader` deps from victory. Nothing depends on them.

PR #1059, opened 2018-07-25
Added react and prop-types to peer dependencies

pastelsky:
Among others, tools like `bundlephobia` (bundlephobia.com) depend on the `peerDependecies` field to be able to determine external dependencies.

### DIFF
--- a/packages/victory/package.json
+++ b/packages/victory/package.json
@@ -40,6 +40,10 @@
     "victory-voronoi-container": "^30.0.0",
     "victory-zoom-container": "^30.0.0"
   },
+  "peerDependencies": {
+    "react": ">=15.0.0",
+    "prop-types": ">=15.0.0"
+  },
   "scripts": {
     "version": "nps build-libs && nps build-dists"
   },


boygirl:
@pastelsky thanks for this PR. I'm going to escalate this question to our CTO.

@ryan-roemer can you weigh in on our decision to _not_ explicitly list `peerDependencies` in `package.json`, and whether that still makes sense.

boygirl:
@ryan-roemer thank you! I will ticket out the `prop-types` change. 

@pastelsky it sounds like we are steering clear of a `peerDependency` on `react`, so I will close this PR. Thank you for helping out!

pastelsky:
Just saw the tickets linked by @ryan-roemer <https://github.com/ryan-roemer>.
While I agree that bumping up peerDependencies every time can be painful,
the field is not strickly enforced by either npm or yarn for all practical
purposes.

Adding a catchall version like "15.x.x" is not adding a guarantee that
victory works with those react versions. Atleast not anymore than not
specifying it at all.

PR #1124, opened 2018-09-23
Change cursor when allowResize and allowDrag are false

erick2014:
This fixes VictoryBrushContainer should not show "rezise-cursor" when allowResize property is false, https://github.com/FormidableLabs/victory/issues/1095

### DIFF
--- a/packages/victory-brush-container/src/victory-brush-container.js
+++ b/packages/victory-brush-container/src/victory-brush-container.js
@@ -83,6 +83,7 @@ export const brushContainerMixin = (base) => class VictoryBrushContainer extends
     const { x, y } = coordinates;
     const { brushStyle, brushComponent, name } = props;
     const brushComponentStyle = brushComponent.props && brushComponent.props.style;
+    const cursor = !props.allowDrag && !props.allowResize ? "pointer" : "move";
     return x[0] !== x[1] && y[0] !== y[1] ?
       React.cloneElement(brushComponent, {
         key: `${name}-brush`,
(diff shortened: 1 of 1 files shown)


boygirl:
@erick2014 thanks for helping out! There are a couple issues with this, though. I'm including a gif to illustrate what I'm talking about:

cursor-issues

1) switching to `cursor: "pointer"` looks strange. I think `cursor: "auto"` is the appropriate value when no resize / move interactions are possible

2) `cursor: "ns-resize"` still shows up. I think you need to alter `yProps` the same way you've altered `xProps`

packages/victory-brush-container/src/victory-brush-container.js
```diff
@@ -102,8 +103,17 @@ export const brushContainerMixin = (base) => class VictoryBrushContainer extends
...
     const style = defaults({}, handleComponentStyle, handleStyle);
-    const yProps = { style, width, height: handleWidth, cursor: "ns-resize" };
-    const xProps = { style, width: handleWidth, height, cursor: "ew-resize" };
+
+    let yProps = { style, width, height: handleWidth, cursor: "ns-resize" };
+    let xProps = { style, width: handleWidth, height, cursor: "ew-resize" };
+
+    if (!props.allowResize && props.allowDrag) {
+      xProps = assign(xProps, { cursor: "move" });
+      yProps = assign(yProps, { cursor: "auto" });
```
boygirl: Should these both be "move" ?

packages/victory-brush-container/src/victory-brush-container.js
```diff
@@ -83,27 +83,46 @@ export const brushContainerMixin = (base) => class VictoryBrushContainer extends
...
   }
 
+  getCursorPointers(props) {
+    const cursors = {
+      "yProps": "ns-resize",
+      "xProps": "ew-resize"
+    };
+    if (!props.allowResize && props.allowDrag) {
+      cursors.xProps = "move";
+      cursors.yProps = "auto";
```
boygirl: I still think these should both be "move". If you disagree can you explain why not?

erick2014: yeah, I agree, because if you can't resize, but you can move, makes sense

boygirl:
🎉 Merging now!

PR #1283, opened 2019-04-01
re-formatted themes to use spread operator instead of _.assign

dsm23:


boygirl:
@dsm23 can you explain the incentive for this change?

It also looks like you're failing CI on some prettier formatting. Would you mind running `nps format` and committing the results?

dsm23:
To reduce the number of lines of code and reduce the dependency on lodash.
I have run the `nps format` command and committed the changes.

boygirl:
@dsm23 `_.assign` is used throughout the codebase, so I don't find this to be a compelling change. Thank you for spending time on Victory anyway!

dsm23:
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

PR #1295, opened 2019-04-16
add diff labels and orientation support

ashika01:
Fixes: #1286 

Improvements on label support for `victory-candlestick`. This PR includes following label support for candlestick 

- lowLabels
- lowLabelComponent
- highLabels
- highLabelComponent
- openLabels
- openLabelComponent
- closeLabels
- closeLabelComponent

There is also continued support for `labels` and `labelComponent`.  `style` can be applied on all the different labels.

Also support `top`,` bottom`, `left`, `right` label orientation support for all the labels

TODO:
- [x] regression testing
- [x] demo update

### DIFF
--- a/packages/victory-candlestick/src/helper-methods.js
+++ b/packages/victory-candlestick/src/helper-methods.js
@@ -224,7 +224,7 @@ const getLabelProps = (dataProps, text, style, type) => {
     scale,
     datum,
     data,
-    orientation: labelOrientation,
+    orientation,
     textAnchor: labelStyle.textAnchor || defaultTextAnchors[orientation],
     verticalAnchor: labelStyle.verticalAnchor || defaultVerticalAnchors[orientation],
     angle: labelStyle.angle,
--- a/packages/victory-candlestick/src/helper-methods.js
+++ b/packages/victory-candlestick/src/helper-methods.js
@@ -180,7 +180,7 @@ const calculatePlotValues = (props) => {
 /* eslint-enable complexity*/
 
 /* eslint-disable max-params*/
-const getLabelProps = (dataProps, text, style, type) => {
+const getLabelProps = (props, text, style, type) => {
   const {
     x,
     high,
@@ -194,7 +194,7 @@ const getLabelProps = (dataProps, text, style, type) => {
     horizontal,
     candleWidth,
     labelOrientation
-  } = dataProps;
+  } = props;
 
   const orientation = getOrientation(labelOrientation, type);
   const positions = { high, low, open, close };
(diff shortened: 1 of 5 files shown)


packages/victory-candlestick/src/helper-methods.js
```diff
@@ -76,34 +113,131 @@ const getDataStyles = (datum, style, props) => {
...
+    return null;
+  } else if (labelProp === true || labels === true) {
+    const dataName = `_${type}`;
+    return `${datum[dataName]}`;
+  }
+  return Array.isArray(labelProp) ? labelProp[index] : labelProp;
+};
+
+const getCandleWidth = (props, style) => {
+  const { active, datum, data, candleWidth, scale, defaultCandleWidth } = props;
```
boygirl: This is a little confusing. You're using the same function that was in `Candle` and supplying it with the data props that you're calculating for each `Candle`. That mostly works, but `active` will not typically be defined at this level. Typically it would get added to state by Victory's event system.

packages/victory-candlestick/src/helper-methods.js
```diff
@@ -76,34 +113,131 @@ const getDataStyles = (datum, style, props) => {
...
+  const candleRatio = props.candleRatio || 0.5;
+  const defaultWidth = candleRatio * (data.length < 2 ? defaultCandleWidth : extent / candles);
+  return Math.max(1, defaultWidth);
+};
+
+const getOrientation = (labelOrientation, type) =>
+  (typeof labelOrientation === "object" && labelOrientation[type]) || labelOrientation;
+
+/* eslint-disable complexity*/
+const calculateAxisValues = (props) => {
```
boygirl: I'm confused about this function name. I don't think axis is accurate here.

boygirl:
@ashika01 I've been trying out this branch, and looking for edge cases. There are a couple issues:

- tooltips don't look right with some labelOrientations. Here's some `openLabel` tooltips with `labelOrientation: "top"`.  
Screen Shot 2019-04-18 at 10 52 12 AM

I'm also seeing a few console warnings in this situation: `Warning: Failed prop type: Invalid prop orientation supplied to VictoryTooltip.`

- It looks like the `Box` / `Border` primitive is trying to render elements with unexpected props:
Screen Shot 2019-04-18 at 10 57 57 AM

I was also not able to alter the new labels via events. For example:
```
events={[{
  target: "data",
  eventHandlers: {
    onMouseOver: () => {
      return [
        {
          target: "openLables",
          mutation: () => {
            return { text: "YO" }
          }
        }
      ];
    }
  }
}]}
```

I'll look into the event issue if you'd prefer. It looks potentially confusing.

ashika01:
Last one include change in box plot as well

Issue #1473, opened 2020-01-21
Tooltip with VoronoiContainer doesn't get multiple datapoints info


jotak:
I need to build some customized tooltips with voronoi containers but with the different options given to me (custom flyout, custom label) I can't find any easy way to get *all* active datapoints information. There's always just a single `datum` object provided to props/callbacks whereas I would expect an array.

We can see in `victory-voronoi-container.js` how this is imho a bit flawed, as in some functions (`getStyle`, getting text in `getLabelProps`) the `points` array is fully reduced (which is good), whereas later on still in `getLabelProps` , `points[0]` is invoked to pick up datum, resulting in providing incomplete information.

If there's no objection I'd like to propose a pull request to add a new `data` or `points` props, passed down to flyout / label. (not changing `datum` to array for obvious compatibility reason)

Or if I'm missing something, if there are some alternative for achieving what I want, it would be kind to tell me. Currently I've got a very hacky workaround by encoding information I need within label text: https://github.com/kiali/kiali-ui/pull/1618

What do you think?

PR #1474, opened 2020-01-21
Add activePoints props to flyout & label components in voronoi container

jotak:
Also add a demo to show what it enables

Fixes https://github.com/FormidableLabs/victory/issues/1473

Notes: I've added a demo, not a story, because there's a comment in stories stating "only add visual test for containers that have visual elements without interaction" which isn't the case here (interactions are needed to trigger tooltip)


jotak:
Capture d’écran de 2020-01-21 10-28-26

boygirl:
@jotak thank you for this straightforward solution. I like it! I'll get a patch out for you tonight.

jotak:
Thanks @boygirl !

jotak:
@boygirl is there a release note about 34.0.0 ? I don't find it, I'm trying to understand what are potential breaking changes against 33.x

boygirl:
The change log has been my source of truth: https://github.com/FormidableLabs/victory/blob/master/CHANGELOG.md#3400-2019-12-20

apologies for any confusion

PR #1479, opened 2020-02-01
Added more typescript demos and typescript linting & formatting

mariano-formidable:
This PR adds in a few more typescript demos, and adds `*.ts` and `*.tsx` files to the formatting and linting pipeline.  

Once this PR is merged into `feature/typescript-defs`, I believe we'll be able to make a follow up PR for `feature/typescript-defs` to merge into `master`, and the infrastructure along with lots of typed coverage can be released!

### Screenshots:
Here's a screenshot showing the current list of typescript'd demos:
image


.eslintrc.js
```diff
@@ -0,0 +1,39 @@
...
+      rules: {
+        "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/explicit-function-return-type": "off",
+        "no-magic-numbers": "off",
+        "import/no-duplicates": "off",
+        "react/no-multi-comp": "off",
+        "react/sort-comp": "off"
+      }
+    }
+  ]
```
mariano-formidable: Added an override so typescript files can be linted using their own rules.

package.json
```diff
@@ -52,7 +52,9 @@
     "@types/lodash": "^4.14.149",
     "@types/react": "^16.9.19",
     "@types/react-dom": "^16.9.5",
-    "babel-eslint": "^6.1.2",
+    "@typescript-eslint/eslint-plugin": "^2.19.0",
+    "@typescript-eslint/parser": "^2.19.0",
+    "babel-eslint": "^10.0.3",
```
mariano-formidable: Had to upgrade `babel-eslint`, `eslint`, and a few others in order to get the `typescript-eslint` packages fully working.

.eslintrc.js
mariano-formidable: Also, I changed it to a `js` file so I could use `__dirname`, and super also, I had to add in the `complexity` and `react/sort-comp` rules so that eslint would pass with the exisiting `.js` files like it did before the package upgrades.

PR #1529, opened 2020-04-24
Task/audit victory chart typerscript props

maddles:
- Audit typescript prop defs in VictoryAxisCharts and add the following properties: children, defaultAxes, defaultPolarAxes, endAngle, prependDefaultAxes
- Move polar to the VictoryCommonProps interface

### DIFF
--- a/packages/victory-chart/src/index.d.ts
+++ b/packages/victory-chart/src/index.d.ts
@@ -9,14 +9,23 @@ import {
   VictoryStyleInterface
 } from "victory-core";
 
+export type AxesType = {
+  independent: React.ReactElement;
+  dependent: React.ReactElement;
+}
+
 export interface VictoryChartProps extends VictoryCommonProps {
+  defaultAxes?: AxesType;
+  defaultPolarAxes?: AxesType;
   categories?: CategoryPropType;
+  children?: React.ReactElement | React.ReactElement[];
   domain?: DomainPropType;
-  domainPadding?: DomainPaddingPropType;
+  endAngle?: number;
   events?: EventPropTypeInterface<string, string | number>[];
   eventKey?: StringOrNumberOrCallback;
   innerRadius?: number;
-  polar?: boolean;
+  prependDefaultAxes: boolean;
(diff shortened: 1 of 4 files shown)


packages/victory-chart/src/index.d.ts
```diff
@@ -9,14 +9,23 @@ import {
   VictoryStyleInterface
 } from "victory-core";
 
+export type AxesType = {
```
kale-stew, line 12: Does this type need to be exported? If you aren't referencing it beyond this file, I think you're ok to just declare the `type`

packages/victory-core/src/index.d.ts
```diff
@@ -540,6 +540,7 @@ export interface VictoryCommonProps {
   minDomain?: number | { x?: number; y?: number };
   name?: string;
   padding?: PaddingProps;
+  polar?: boolean;
```
kale-stew, line 543: I can see that you've removed `polar` from a few of the interfaces that already extend `VictoryCommonProps` to reduce redundancy, but would you mind removing it in another spot? I see it in `VictoryCandleStickProps` at first glance, but there might be more duplicity elsewhere.

kale-stew: (I searched for `polar?` in the codebase to pull up duplicate instances 🙂)

maddles: Yep, moved in in https://github.com/FormidableLabs/victory/pull/1529. Was just going to merge that first so I don't have surprise conflicts

packages/victory-chart/src/index.d.ts
maddles: I noticed there was a pattern of exporting types and interfaces from the d.ts files? It's been useful to be able to import types directly in demos

packages/victory-core/src/index.d.ts
maddles: Oh heck, I misinterpreted this. Soz, yes, will audit all of them and pare them down!

PR #1530, opened 2020-04-24
Add types: VictorySelectionContainer

kale-stew:
This PR
- enhances the shared VictoryContainerProps
- adds types for the VictorySelectionContainer
- adds a tsx demo to test the VictorySelectionContainer types

### DIFF
--- a/packages/victory-core/src/index.d.ts
+++ b/packages/victory-core/src/index.d.ts
@@ -220,13 +220,23 @@ export class VictoryLabel extends React.Component<VictoryLabelProps, any> {}
 // #region Victory Container
 
 export interface VictoryContainerProps {
+  children?: React.ReactElement | React.ReactElement[];
+  className?: string;
+  containerId?: number | string;
+  desc?: string;
+  events?: React.DOMAttributes<any>;
+  height?: number;
+  name?: string;
+  origin?: OriginType;
+  polar?: boolean;
+  portalComponent?: React.ReactElement;
+  portalZIndex?: number;
   responsive?: boolean;
   style?: React.CSSProperties;
-  height?: number;
-  width?: number;
-  events?: React.DOMAttributes<any>;
+  tabIndex?: number;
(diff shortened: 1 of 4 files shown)


packages/victory-core/src/index.d.ts
```diff
@@ -218,13 +218,23 @@ export class VictoryLabel extends React.Component<VictoryLabelProps, any> {}
 // #region Victory Container
 
 export interface VictoryContainerProps {
+  children?: React.ReactElement | React.ReactElement[];
```
maddles, line 221: One thing I ran into is if you have a child iterator (i.e. if you use map()), React.ReactElement will fail to validate, but React.ReactNode works

packages/victory-selection-container/src/index.d.ts
```diff
@@ -0,0 +1,19 @@
+import * as React from "react";
+import { VictoryContainerProps } from "victory-core";
+
+interface VictorySelectionContainerProps extends VictoryContainerProps {
+  activateSelectedData?: boolean;
+  allowSelection?: boolean;
+  disable?: boolean;
+  onSelection?: Function;
```
maddles, line 8: Since we know the types for the args here, I think it's possible to go more granular, e.g.:

```
interface PointsInterface = {
childName?: string | string[];
eventKey?: string | number; 
data?: any;
}

{(points: PointsInterface[], bounds: {x: number | Date, y: number | Date}[], props: any) => void}
```

(I don't know that these are the full set of types for these args, but the gist of them: https://formidable.com/open-source/victory/docs/victory-selection-container#onselection)

packages/victory-selection-container/src/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+  onSelectionCleared?: Function;
```
maddles, line 9: Same idea here, e.g.:

`{(props: VictorySelectionProps) => void}`

PR #1531, opened 2020-04-24
Victory Histogram

NgoKnows:
## What it Does
This PR adds a VictoryHistogram component/package, which will allow users to create histogram graphs.

## Screenshot(s)
Screen Shot 2020-05-07 at 11 38 20 AM


packages/victory-bar/src/helper-methods.js
```diff
@@ -1,5 +1,5 @@
 import { assign, isNil } from "lodash";
-import { Helpers, LabelHelpers, Data, Domain, Scale, Collection } from "victory-core";
+import { Helpers, LabelHelpers, Data, Domain, Scale, Collection } from "../../victory-core/src";
```
boygirl, line 2: TODO: change this back to importing from victory-core before merging

packages/victory-bar/src/victory-bar.js
```diff
@@ -11,7 +11,7 @@ import {
   addEvents,
   Data,
   Domain
-} from "victory-core";
+} from "../../victory-core/src";
```
boygirl, line 14: TODO: change this back to importing from victory-core before merging

packages/victory-chart/src/helper-methods.js
```diff
@@ -2,7 +2,7 @@
 /* eslint-disable no-use-before-define */
 
 import React from "react";
-import { Helpers, Scale, Axis, Wrapper } from "victory-core";
+import { Helpers, Scale, Axis, Wrapper } from "../../victory-core/src";
```
boygirl, line 5: ditto

packages/victory-core/src/victory-theme/grayscale.js
```diff
@@ -149,6 +149,19 @@ export default {
     },
     baseProps
   ),
+  histogram: assign(
+    {
+      style: {
+        data: {
+          fill: "palevioletred",
```
boygirl, line 156: grayscale theme should use a grayscale color for the default fill

packages/victory-histogram/src/helper-methods.js
```diff
@@ -0,0 +1,303 @@
...
+    return value;
+  };
+};
+
+const getBinningFunc = ({ data, x, bins }) => {
+  const xAccessor = Helpers.createAccessor(x || "x");
+  const dataIsDates = data.some((datum) => xAccessor(datum) instanceof Date);
+  const bin = d3Array.bin().value(xAccessor);
+  const niceScale = d3Scale
+    .scaleLinear()
```
boygirl, line 77: I think this should use the x domain / scale passed in or calculated from props

boygirl: Actually, nevermind. I understand that this domain is only used for calculating the bin edges, so it makes sense for it to be scoped to just the data / bins of this component.

boygirl: If you allow time scales as well as linear scales, can you get around needing to use "day" "month" "year" bins?

test/client/spec/victory-bar/victory-bar.spec.js
```diff
@@ -166,7 +165,8 @@ describe("components/victory-bar", () => {
           ]}
         />
       );
-      const Labels = wrapper.find(VictoryLabel);
+      const Labels = wrapper.find("VictoryLabel");
+      expect(Labels).to.have.lengthOf(3);
```
NgoKnows, line 169: for some reason this test was previously not matching any nodes, so this fixes that + makes sure we match some nodes

packages/victory-histogram/src/helper-methods.js
```diff
@@ -0,0 +1,247 @@
...
+  }
+  const dataOrBinsContainsDates = dataOrBinsContainDates({ data, bins, x });
+  const binFunc = getBinningFunc({ data, x, bins, dataOrBinsContainsDates });
+  const binnedData = binFunc(data).filter(({ x0, x1 }) => x0 !== x1);
+
+  const formattedData = binnedData.map((bin) => ({
+    x: dataOrBinsContainsDates ? new Date(bin.x0) : bin.x0,
+    end: dataOrBinsContainsDates ? new Date(bin.x1) : bin.x1,
+    y: bin.length,
+    binnedDatums: [...bin]
```
NgoKnows, line 81: what do you think about the name of these properties for the shape of the binned data?

packages/victory-histogram/src/victory-histogram.js
```diff
@@ -0,0 +1,113 @@
...
+      duration: 500,
+      before: () => ({ y: 0 }),
+      after: (datum) => ({ _y: datum._y, _y1: datum._y1, _y0: datum._y0 })
+    }
+  };
+
+  static propTypes = {
+    ...CommonProps.baseProps,
+    ...CommonProps.dataProps,
+    barSpacing: CustomPropTypes.nonNegative,
```
NgoKnows, line 50: `barSpacing` vs `binSpacing`?

boygirl: I like `binSpacing`

packages/victory-histogram/src/helper-methods.js
boygirl: Maybe either `x0` and `x1` or `binEdges` where the value is a tuple. And `binnedData` rather than `binnedDatums`

boygirl:
@NgoKnows it looks like something might be off with transitioning date data:
Screen Shot 2020-05-07 at 3 06 06 PM

packages/victory-histogram/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "main": "lib/index.js",
+  "module": "es/index.js",
+  "jsnext:main": "es/index.js",
+  "author": "Formidable",
+  "license": "MIT",
+  "dependencies": {
+    "d3-array": "^2.4.0",
+    "lodash": "^4.17.15",
+    "prop-types": "^15.5.8",
+    "victory-core": "^34.1.3"
```
boygirl, line 25: `bar` and `path-helper-methods` should be moved to `victory-core/victory-primitives` so that `victory-histogram doesn't need to depend on `victory-bar`

packages/victory-histogram/src/helper-methods.js
```diff
@@ -0,0 +1,245 @@
...
+export const getFormattedData = cacheLastValue(({ data = [], x, bins }) => {
+  if ((!data || !data.length) && !Array.isArray(bins)) {
+    return [];
+  }
+  const dataOrBinsContainsDates = dataOrBinsContainDates({ data, bins, x });
+  const binFunc = getBinningFunc({ data, x, bins, dataOrBinsContainsDates });
+  const binnedData = binFunc(data).filter(({ x0, x1 }) => x0 !== x1);
+
+  const formattedData = binnedData.map((bin) => ({
+    x0: dataOrBinsContainsDates ? new Date(bin.x0) : bin.x0,
```
NgoKnows, line 78: using _x0 and _x1 was causing positioning issues, so I went with x0 and x1

packages/victory-core/src/victory-util/wrapper.js
```diff
@@ -20,6 +20,7 @@ import Events from "./events";
 import Collection from "./collection";
 import Helpers from "./helpers";
 import Scale from "./scale";
+import { getFormattedData } from "./../../../victory-histogram/src/helper-methods";
```
NgoKnows, line 23: So I'm a bit unsure about how to best handle this. I could pull this function out into victory-core, but I would have to pull out all of the functions related to this, so that seems not great. I was thinking of putting a static method on VictoryHistogram that we could reference below instead of directly importing this function.

packages/victory-core/src/victory-util/wrapper.js
```diff
@@ -204,6 +208,42 @@ export default {
     const { polar, startAngle, endAngle, categories, minDomain, maxDomain } = props;
     const parentProps = { polar, startAngle, endAngle, categories, minDomain, maxDomain };
     let stack = 0;
+    const children = childComponents
+      ? childComponents.slice(0)
+      : React.Children.toArray(props.children);
+
+    const childrenAreHistograms = children.some((child) => {
```
NgoKnows, line 215: is this check sufficient?

NgoKnows:
Regarding the date animation issues, I looked into it, and if you are referring to the jumpiness that happens when new bins are animated in. I found that it was when putting `animation={{ duration: 100 }}` on VictoryChart this happens. However it looks to be a pre-existing issue, as I tried this with VictoryBar and the same behavior happens.

packages/victory-core/src/victory-util/wrapper.js
boygirl: Yeah, I'd definitely prefer it as a static on `VictoryHistogram`

boygirl:
The date animation issue I'm referring to, is the ~half of a bin at the end of the set. It looks like a domain issue, maybe? Or maybe bins being calculated based on stale data

packages/victory-core/src/victory-util/wrapper.js
boygirl: Since stacking histograms will be incompatible with other types, I think we need to check on  whether `.some`children are histograms,  and then warn if not `.every` child is a histogram.

stories/victory-histogram.js
```diff
@@ -0,0 +1,955 @@
...
+  { x: 56 },
+  { x: 18 },
+  { x: 18 },
+  { x: 28 },
+  { x: 79 },
+  { x: 44 }
+];
+
+const timeData = [
+  { x: new Date("2018-09-24T07:00:00.000Z") },
```
boygirl, line 61: Can you move this to a new file to make the stories a bit easier to read through?

stories/victory-histogram.js
```diff
@@ -0,0 +1,955 @@
...
+  ]);
+
+const stackedData = [
+  ...[50, 30, 100, 32, 50, 10, 49, 78, 20].map((count) => getData(count, count, 100)),
+  [{ x: 1 }, { x: 3 }, { x: 1 }, { x: 2 }]
+];
+
+storiesOf("VictoryHistogram.stacked", module)
+  .addDecorator(getChartDecorator())
+  .add("stacked area with explicit bins passed to stack", () => (
```
boygirl, line 920: `area` -> `histogram`

stories/victory-histogram.js
```diff
@@ -0,0 +1,955 @@
...
+    </VictoryStack>
+  ))
+  .add("stacked area with numeric bins passed to stack", () => (
+    <VictoryStack colorScale="qualitative" bins={2}>
+      {stackedData.map((d, index) => (
+        <VictoryHistogram data={d} key={index} />
+      ))}
+    </VictoryStack>
+  ))
+  .add("stacked area with bins passed to first child", () => (
```
boygirl, line 934: based in the title of this story, I would expect a `bins` prop on the first histogram child

stories/victory-histogram.js
```diff
@@ -0,0 +1,955 @@
...
+
+const stackedData = [
+  ...[50, 30, 100, 32, 50, 10, 49, 78, 20].map((count) => getData(count, count, 100)),
+  [{ x: 1 }, { x: 3 }, { x: 1 }, { x: 2 }]
+];
+
+storiesOf("VictoryHistogram.stacked", module)
+  .addDecorator(getChartDecorator())
+  .add("stacked area with explicit bins passed to stack", () => (
+    <VictoryStack colorScale="qualitative" bins={[0, 20, 100]}>
```
boygirl, line 921: please add `bins` to `VictoryStack` `propTypes`

packages/victory-histogram/src/helper-methods.js
boygirl: Is this because of `Helpers.scalePoint` in `Bar`?

packages/victory-voronoi/src/index.d.ts
```diff
@@ -13,7 +13,7 @@ import {
 export type VictoryVoronoiSortOrderType = "ascending" | "descending";
 
 export interface VictoryVoronoiProps
-  extends VictoryCommonProps,
+  extends Omit<VictoryCommonProps, "polar">,
```
NgoKnows, line 16: removed the polar prop type from components that currently don't support it

packages/victory-core/src/victory-util/wrapper.js
```diff
@@ -204,6 +208,50 @@ export default {
...
     let stack = 0;
+    const children = childComponents
+      ? childComponents.slice(0)
+      : React.Children.toArray(props.children);
+
+    const someChildrenAreHistograms = children.some((child) => {
+      return child.type && child.type.role === "histogram";
+    });
+
+    const allChildrenAreHistograms =
```
boygirl, line 219: We should only perform this check if `someChildrenAreHistograms`

packages/victory-core/src/victory-util/wrapper.js
```diff
@@ -204,6 +208,50 @@ export default {
...
+    if (someChildrenAreHistograms && !allChildrenAreHistograms) {
+      Log.warn(
+        "VictoryHistogram only supports being stacked with other VictoryHistogram components. Check to make sure that you are only passing VictoryHistogram components to VictoryStack"
+      );
+    }
+
+    // if we are stacking histograms, we need to generate explicit bins
+    // or else each histogram may end up having different bins
+    if (allChildrenAreHistograms) {
+      let childBins = props.bins || childComponents[0].props.bins;
```
boygirl, line 234: I think it would be more readable to pull all the bin logic into a separate function

packages/victory-stack/src/victory-stack.js
```diff
@@ -19,6 +26,12 @@ export default class VictoryStack extends React.Component {
 
   static propTypes = {
     ...CommonProps.baseProps,
+    bins: PropTypes.oneOfType([
```
boygirl, line 29: This should also be added to the TS types

boygirl:
I'm seeing some transition oddities that don't happen on `VictoryBar`. These are happening on both time series and regular histograms
odd-transition

packages/victory-histogram/src/victory-histogram.js
```diff
@@ -0,0 +1,115 @@
...
+      before: () => ({ _y: 0, _y1: 0, _y0: 0 }),
+      after: (datum) => ({ _y: datum._y, _y1: datum._y1, _y0: datum._y0 })
+    },
+    onExit: {
+      duration: 500,
+      before: () => ({ _y: 0, yOffset: 0 })
+    },
+    onEnter: {
+      duration: 500,
+      before: () => ({ y: 0 }),
```
boygirl, line 42: In `VictoryBar` the `onEnter` transition uses ` before: () => ({ _y: 0, _y1: 0, _y0: 0 })`. Maybe changing this would correct the entrance issue in histograms

NgoKnows: good catch, fixed this!

boygirl:
@NgoKnows This is looking awesome and complete! Approved!

PR #1536, opened 2020-04-25
Audits and Adds typescript props, definitions, and demo: VictoryZoomContainer

wsparsons:
* Updates the name of `CursorData` to `CoordinatesPropType` 
* Updates the `VictoryBrushContainerProps` and `VictoryBrushContainerDemo` to use `DomainTuple` instead `RangeTuple`
* Adds a `VictoryZoomContainerDemo`
* Adds and updates missing props and their typescript definitions for `VictoryZoomContainerProps` interface
* Removes props from `VictoryZoomContainerProps` interface that do not belong

### DIFF
--- a/demo/js/components/victory-zoom-container-demo.js
+++ b/demo/js/components/victory-zoom-container-demo.js
@@ -2,7 +2,7 @@
 /*eslint-disable no-magic-numbers,react/no-multi-comp */
 import React from "react";
 import PropTypes from "prop-types";
-import { range, merge, random, minBy, maxBy, last, round } from "lodash";
+import { range, merge, random, minBy, maxBy, last } from "lodash";
 import { VictoryChart } from "Packages/victory-chart/src/index";
 import { VictoryStack } from "Packages/victory-stack/src/index";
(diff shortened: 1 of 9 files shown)


demo/js/components/victory-zoom-container-demo.js
```diff
@@ -123,7 +118,7 @@ export default class App extends React.Component {
 
   getZoomDomain() {
     return {
-      y: [random(0, 0.4, 0.1), random(0.6, 1, 0.1)]
```
wsparsons, line 126: ended up removing this since https://lodash.com/docs/4.17.15#random says the third argument should be a boolean. not sure if the `0.1` was intentional or if it should have a boolean value.

demo/js/components/victory-zoom-container-demo.js
```diff
@@ -61,11 +61,6 @@ class CustomChart extends React.Component {
       x: [data[0].x, last(data).x]
     };
   }
-  getZoomFactor() {
```
wsparsons, line 64: ended up removing this uninvoked function. not sure if this should be used in the demo and somehow got missed?

PR #1551, opened 2020-04-30
Add draggable demo, make sure DomainPropType is used where approriate

maddles:
This PR:

- Adds the draggable demo
- Replaces object literals with DomainPropType where appropriate

### DIFF
--- a/packages/victory-core/src/index.d.ts
+++ b/packages/victory-core/src/index.d.ts
@@ -621,7 +621,7 @@ export namespace Selection {
     evt: React.SyntheticEvent,
     svg?: SVGElement
   ): SVGCoordinateType;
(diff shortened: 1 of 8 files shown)


packages/victory-core/src/index.d.ts
```diff
@@ -461,10 +461,10 @@ export interface EventPropTypeInterface<TTarget, TEventKey> {
 }
 
 export type DomainTuple = [number, number] | [Date, Date];
-export type DomainPropType =
-  | DomainTuple
+export type DomainPropObjectType =
   | { x?: DomainTuple; y: DomainTuple }
   | { x: DomainTuple; y?: DomainTuple };
+export type DomainPropType = DomainPropObjectType | DomainTuple;
```
wsparsons, line 467: thank you for extracting this type out. it will make it easier to use the different domain types

PR #1580, opened 2020-05-18
Fix Histogram Tooltip Positioning

NgoKnows:
This fixes an issue that was occurring where tooltips being rendered by `VictoryVoronoiContainer`  for `VictoryHistogram` were being placed at the beginning of the bin (`x0`), which leads to a slightly weird positioning. To fix that, the middle of the bin is now used, so this should lead to a more intuitive positioning. 

Screen Shot 2020-05-18 at 11 29 29 AM

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,7 @@ before_install:
 script:
 - yarn --version
 - yarn nps check.ci
-# disable chromatic until billing issues are resolved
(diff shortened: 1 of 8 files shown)


boygirl:
It looks like this change alters how bin spacing is applied. It's no longer symmetric, but applied to the beginning of each bin. Is this intended?

Screen Shot 2020-05-18 at 1 47 50 PM

NgoKnows:
Ah woops, no it was not intended. It was due to changing alignment to "middle', but I've fixed up the bar offset logic, so hopefully that should fix things.

boygirl:
Approved! Thank you!

PR #1583, opened 2020-05-19
Get background rendering in labels

maddles:
This pull request adds backgrounds to VictoryLabel. 

These backgrounds can be:

1. One single big background for the entire label, as added with a single object for the new `backgroundStyle` prop on VictoryLabel, e.g. `backgroundStyle={{ fill: "pink" }}`
2. One background for each `<tspan>` within a given label when the `backgroundStyle` prop is passed as an array, e.g. `backgroundStyle={[ { fill: "pink" },  { fill: "blue" }]}`

This also adds the prop `backgroundPadding`, which takes the standard Victory `padding` prop (a number, an object, or an array of objects) and will add padding to your background or backgrounds.

Another added prop, backgroundComponent, uses the Victory `Rect` primitive component by default, but can be optionally set explicitly on VictoryLabel.

Screenshots!

Screenshot 2020-06-12 at 6 01 25 PM
Screenshot 2020-06-12 at 6 01 44 PM


packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -1,14 +1,16 @@
 import React from "react";
 import PropTypes from "prop-types";
 import VictoryPortal from "../victory-portal/victory-portal";
+import Background from "../victory-primitives/background";
```
boygirl, line 4: Since we never want to render circular backgrounds here, maybe we can get away with using `Rect` for the background component

wsparsons: I will investigate about seeing we can just use `Rect` instead of `Background`

packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -110,36 +115,225 @@ const getTransform = (props) => {
...
 };
 
-const renderElements = (props) => {
-  const { inline, className, title, events, direction, text, style } = props;
-  const lineHeight = getHeight(props, "lineHeight");
-  const textAnchor = props.textAnchor ? Helpers.evaluateProp(props.textAnchor, props) : "start";
-  const dx = props.dx ? Helpers.evaluateProp(props.dx, props) : 0;
-  const dy = getDy(props, lineHeight);
+const getBackgroundPadding = (padding) => {
+  return typeof padding === "object" && padding !== null
```
boygirl, line 119: You can probably reuse logic from `Helpers.getPadding` here.

packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -110,36 +115,225 @@ const getTransform = (props) => {
...
+
+  return React.cloneElement(
+    backgroundComponent,
+    defaults({}, backgroundComponent.props, backgroundProps)
+  );
+};
+
+const getPaddingProp = (i, padding) => {
+  if (Array.isArray(padding)) {
+    if (padding[i]) {
```
boygirl, line 205: If padding is an array, let's have it use `padding[0]` if `padding[i]` doesn't exist to be consistent with how style arrays work

packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -110,36 +115,225 @@ const getTransform = (props) => {
...
+          : dy - totalLineHeight * 0.5 - (currentStyle.fontSize - capHeightPx)
+    };
+  });
+
+  const backgroundStyleChildren = backgroundStyle.map((bgStyle, i) => {
+    const xCoordinate = getXCoordinate(calculatedProps, textElement[i].labelSize.width);
+    const yCoordinate = textElement.slice(0, i + 1).reduce((prev, curr) => {
+      return prev + curr.dy;
+    }, y);
+    const transform = getTransform(props);
```
boygirl, line 249: can `transform` come from `calculatedProps` rather than being recalculated?

packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -110,36 +115,225 @@ const getTransform = (props) => {
...
+  const capHeightsPx = TextSize.convertLengthToPixels(`${capHeight}em`, maxFontSize);
+  const textHeight = inline
+    ? maxFontSize * adjustedLineHeight + capHeightsPx
+    : getBlockTextHeight(props, adjustedLineHeight, capHeightsPx);
+  const width = inline
+    ? TextSize.approximateTextSize(text.join(" "), style).width + (dx || 0) * text.length
+    : TextSize.approximateTextSize(maxString, style).width + (dx || 0);
+  const xCoordinate = getXCoordinate(calculatedProps, width);
+  const yCoordinate = getYCoordinate(calculatedProps, props, textHeight);
   const transform = getTransform(props);
```
boygirl, line 186: This `transform` too

packages/victory-core/src/index.d.ts
```diff
@@ -193,6 +193,9 @@ export type LabelOrientationType = "parallel" | "perpendicular" | "vertical";
 
 export interface VictoryLabelProps {
   angle?: string | number;
+  backgroundComponent?: React.ReactElement;
+  backgroundStyle?: React.CSSProperties;
+  backgroundPadding?: PaddingProps | PaddingProps[];
```
boygirl, line 198: Should also have `groupComponent` now, I think

packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -172,12 +366,56 @@ const evaluateProps = (props) => {
   return assign({}, props, { style, text, id });
 };
 
+const getPadding = (props) => {
```
boygirl, line 369: Can this be renamed `getBackgroundPadding` for clarity?

stories/victory-label.js
```diff
@@ -0,0 +1,174 @@
...
+        {story()}
+      </svg>
+    );
+  };
+};
+
+storiesOf("VictoryLabel.background", module)
+  .addDecorator(svgDecorator())
+  .add("single background on multiple lines", () => (
+    <VictoryLabel
```
boygirl, line 19: Please add stories that include `backgroundPadding`

packages/victory-polar-axis/src/victory-polar-axis.js
```diff
@@ -97,7 +97,8 @@ class VictoryPolarAxis extends React.Component {
       axisLabel: PropTypes.object,
       grid: PropTypes.object,
       ticks: PropTypes.object,
-      tickLabels: PropTypes.object
+      tickLabels: PropTypes.object,
+      background: PropTypes.object
```
boygirl, line 101: I don't think this prop should be added here.

boygirl:
It looks like variable line heights aren't correctly taken into account with single backgrounds: For example:

```
<VictoryLabel
  backgroundStyle={{ fill: "pink" }}
  style={[
    { fontSize: 40, fill: "green" },
    { fontSize: 10 }
  ]}
  lineHeight={[1, 3, 1]}
  x={300}
  y={450}
  textAnchor="start"
  verticalAnchor="start"
  text={"Victory is awesome.\nThis is (start, start) anchoring.\nCapisce?"}
/>
```

Screen Shot 2020-06-13 at 1 18 03 PM

boygirl:
I'm working on a bit of refactoring in a separate branch. I'll have that up for you before Monday morning.

packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -110,36 +115,225 @@ const getTransform = (props) => {
...
+
+const getTSpanDy = (props, calculatedProps, i) => {
+  const { backgroundStyle, inline } = props;
+  const { padding } = calculatedProps;
+  const paddingProp = getPaddingProp(i, padding);
+  const prevPaddingProp = getPaddingProp(i - 1, padding);
+
+  if (i && !inline) {
+    return backgroundStyle && Array.isArray(backgroundStyle) && backgroundStyle.length > 1
+      ? calculateSpanDy(props, calculatedProps, i) + prevPaddingProp.top + prevPaddingProp.bottom
```
boygirl, line 308: I don't know that the background padding should influence the text positioning

boygirl:
It looks like there is a `lineHeight` regression on this branch as well:

this branch:
Screen Shot 2020-06-13 at 1 41 25 PM

on master:
Screen Shot 2020-06-13 at 1 41 01 PM

packages/victory-core/src/victory-label/victory-label.js
maddles: This is a frustrating one because it has to for the multi-background, but there is a condition there to make sure it's only applied when there are actually multiple background styles

boygirl: No, I think the background padding explicitly _should not_ change the text positioning. I understand that means that when padding is applied, background rects might not exactly line up with one another. But, the alternative is that capHeight / lineHeight are non-deterministic for labels when there is a backgroundPadding prop.

packages/victory-core/src/victory-util/helpers.js
```diff
@@ -90,6 +90,13 @@ function getPadding(props) {
   };
 }
 
+function getBackgroundPadding(props) {
+  const { backgroundPadding } = props;
+  const padding = backgroundPadding;
+
+  return getPadding({ padding });
+}
+
```
boygirl, line 99: Instead of this, what about altering `getPadding` like so:

```
function getPadding(props, name = "padding") {
  const padding = props[name];
  const paddingVal = typeof padding === "number" ? padding : 0;
  const paddingObj = typeof padding === "object" ? padding : {};
  return {
    top: paddingObj.top || paddingVal,
    bottom: paddingObj.bottom || paddingVal,
    left: paddingObj.left || paddingVal,
    right: paddingObj.right || paddingVal
  };
}
```

packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -17,6 +20,8 @@ const defaultStyles = {
   stroke: "transparent"
 };
 
+const defaultPadding = { top: 0, left: 0, bottom: 0, right: 0 };
```
boygirl, line 23: This shouldn't be necessary if using `Helpers.getPadding`

packages/victory-core/src/victory-label/victory-label.js
boygirl: Screen Shot 2020-06-15 at 9 44 22 PM

vs

Screen Shot 2020-06-15 at 9 45 08 PM

boygirl: I think it's more correct to maintain the spacing between text elements than to make the background line up, but maybe this is a matter of preference.

boygirl:
@maddles this PR is approved! Great work! This will be released as a minor version later this week. I'll take care of the `victory-native` changes and docs updates.

PR #1603, opened 2020-06-05
Update tspan dy calculation

NgoKnows:


packages/victory-core/src/victory-label/victory-label.js
```diff
@@ -213,9 +213,8 @@ const getChildBackgrounds = (props, calculatedProps) => {
       fontSize: style.fontSize || defaultStyles.fontSize,
       dy:
         i && !inline
-          ? Math.ceil(previousStyle.fontSize * previousLineHeight) + Math.ceil(capHeight * 2)
-          : Math.ceil(((adjustedLineHeight * currentStyle.fontSize) % currentStyle.fontSize) / 2) +
-            Math.ceil(capHeight)
+          ? Math.ceil(previousStyle.fontSize * previousLineHeight)
+          : Math.ceil(((adjustedLineHeight * currentStyle.fontSize) % currentStyle.fontSize) / 2)
```
NgoKnows, line 217: I think this modulo is causing issues with decimal line heights

PR #1705, opened 2020-10-06
Feature/accessibility group component

ljones87:
- Adds an `AccessibilityGroupComponent` to `victory-core`. 
- This component can be provided to the `groupComponent` prop to add `aria-labels` and `desc` tags with `aria-describedby`.
- Also adds `aria-label` to `groupComponent` option `VictoryClipContainer`
- [x] tests

AccessibilityGroupComponent

### DIFF
--- a/packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
+++ b/packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
@@ -1,0 +1,42 @@
+import React from "react";
+import PropTypes from "prop-types";
+// import CustomPropTypes from "../victory-util/prop-types";
+// import Helpers from "../victory-util/helpers";
+import { assign /*defaults, isObject, uniqueId */ } from "lodash";
+
+class VictoryAccessibilityGroup extends React.Component {
+  static displayName = "VictoryAccessibilityGroup";
+  static role = "container";
+  static propTypes = {
+    "aria-describedby": PropTypes.string,
+    "aria-label": PropTypes.string,
+    desc: PropTypes.string,
+    descId: PropTypes.string,
+    children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node]),
+    className: PropTypes.string,
+    tabIndex: PropTypes.number
+  };
+
+  static defaultProps = {
+    className: "VictoryAccessibilityGroup"
+  };
+
+  render() {
+    const { desc, children, className, descId, tabIndex } = this.props;
+
+    return desc ? (
+      <g aria-label={this.props["aria-label"]} className={className} tabIndex={tabIndex}>
+        <desc aria-describedby={this.props["aria-describedby"]} id={descId}>
+          {desc}
+        </desc>
+        {children}
+      </g>
+    ) : (
+      <g aria-label={this.props["aria-label"]} className={className} tabIndex={tabIndex}>
+        {children}
+      </g>
+    );
+  }
+}
+
+export default VictoryAccessibilityGroup;
--- a/packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
+++ b/packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
@@ -1,8 +1,5 @@
 import React from "react";
 import PropTypes from "prop-types";
-// import CustomPropTypes from "../victory-util/prop-types";
-// import Helpers from "../victory-util/helpers";
-import { assign /*defaults, isObject, uniqueId */ } from "lodash";
 
 class VictoryAccessibilityGroup extends React.Component {
   static displayName = "VictoryAccessibilityGroup";
--- a/demo/demo-data.ts
+++ b/demo/demo-data.ts
@@ -91,3 +91,22 @@ export const accessibilityErrorBarData = [
   { x: 34, y: 65, error: 0.15 },
   { x: 43, y: 50, error: 0.2 }
 ];
+
+export const accessibilityGroupData = {
+  a: [
+    { x: 1, y: 1 },
+    { x: 2, y: 2 },
+    { x: 3, y: 5 }
+  ],
+  b: [
+    { x: 1, y: 2 },
+    { x: 2, y: 1 },
+    { x: 3, y: 7 }
+  ],
+  c: [
+    { x: 1, y: 3 },
+    { x: 2, y: 4 },
+    { x: 3, y: 9 }
+  ]
+
+}
--- a/demo/demo-data.ts
+++ b/demo/demo-data.ts
@@ -108,5 +108,4 @@ export const accessibilityGroupData = {
     { x: 2, y: 4 },
     { x: 3, y: 9 }
   ]
-
-}
+};
--- a/demo/js/components/accessibility-demo.js
+++ b/demo/js/components/accessibility-demo.js
@@ -58,7 +58,8 @@ const chartContainerStyle = {
   flexDirection: "column",
   alignItems: "center",
   width: "50%",
-  height: "50%"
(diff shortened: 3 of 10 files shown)


packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+class VictoryAccessibilityGroup extends React.Component {
```
ljones87, line 4: preference of Victory _Accessible_ Group vs _Accessibility_ ?

boygirl: I think `VictoryAccessibleGroup` reads a little nicer, but I don't have a strong preference.

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
```diff
@@ -0,0 +1,39 @@
...
+  static defaultProps = {
+    className: "VictoryAccessibilityGroup"
+  };
+
+  render() {
+    const { desc, children, className, descId, tabIndex } = this.props;
+
+    return desc ? (
+      <g aria-label={this.props["aria-label"]} className={className} tabIndex={tabIndex}>
+        <desc aria-describedby={this.props["aria-describedby"]} id={descId}>
```
boygirl, line 26: I think `aria-describedby` belongs on the `g` tag. When there is a `desc`, `aria-describedby` should use the id of the `desc` tag. When there isn't a `desc`, it should use whatever user provided id string.

boygirl, line 26: Also, when there is a `desc`, there needs to be a back up for if the user doesn't provide a `descId`

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+class VictoryAccessibilityGroup extends React.Component {
+  static displayName = "VictoryAccessibilityGroup";
+  static role = "container";
```
boygirl, line 6: Since you've added `role = "container"`, this component will have access to a bunch of additional props added by its parent when it's used on a component like `VictoryBar` rather than on one like `VictoryGroup` is in your demo. This should let you come up with a descriptive id more easily when you need one.

boygirl: To have it work identically with `VictoryGroup`, `VictoryStack`,  and `VictoryChart` though, some additional work would be needed on those components so ensure they're passing the same props to their groupComponent that they pass to their `containerComponent` if it has `role = "container"`

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
ljones87: ok cool, yeah I didn't get a lot out of looking up aria-describedby best practice, but in looking up reading on the `<desc>` tag i see its placement on `<desc>` is redundant. Will fix

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
ljones87: gotcha, is that the preference or should the role prop be removed?

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
ljones87: What about enforcing this with types?
No error if `desc` is not provided (will just render the `g` w/props)
fine

Then if `desc` is present, require `descId`
fine

I'm sure theres a more efficient way to go about this, but basically:
```

export type VictoryAccessibleGroupProps = {
  desc: string;
  descId: string;
  "aria-describedby"?: string;
  "aria-label": string;
  children?: React.ReactElement | React.ReactElement[];
  className?: string;
  tabIndex?: number;
} | {
  desc?: false;
  descId?: never;
  "aria-describedby"?: string;
  "aria-label": string;
  children?: React.ReactElement | React.ReactElement[];
  className?: string;
  tabIndex?: number;
}
```

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
boygirl: It depends on what you end up doing as a backup `descId`. I think they only thing you might want from those props would be the key / id that Victory generates for children of a chart that don't have explicit names. You'd end up with something like `chart-group-0-bar-1-parent` for example, which would be useful. If you end up going with a different solution for `descId`, and don't need those props, you should ditch the `container` role.

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
ljones87: OK found a better resource to understand the described by attribute, https://developer.paciellogroup.com/blog/2018/09/describing-aria-describedby/

in this article, `aria-describedby` and `descId` should be identical, with `aria-describedby` using the `descId` (as you had said)
I am wondering if `aria-describedby` should even be a prop the user controls: 
```
if(descId) => descId is required, aria-describedby will use the same input
else arial-label should suffice
```
error

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
ljones87: I see, so I could allow `descId` unrequired `if(desc)` and generate it from children. Cool I will look at this now

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
boygirl: Hmm... I think `aria-describedby` should still be a prop in the case where a user wants to describe the group with some other, non-victory element. If they provide both a `desc` prop, though, the user's `aria-describedby` should be used to set the `descId` so that they definitely match. If they provide only a `desc` and no `aria-describedby` we will need to generate an id to use for both `aria-describedby` and `descId`. So, I think `descId` probably shouldn't be a prop.

packages/victory-core/src/victory-accessibility-group/victory-accessibility-group.js
boygirl: Yeah, you might try adding `VictoryAccessibleGroup` to something like `VictoryBar` to see what you're working with. I can help you figure out how to get `VictoryGroup`, `VictoryStack` and `VictoryChart` to also pass these props to their `container` `groupComponents`

ljones87: Would something this work or should we add/pass a `name` (or `id`/preferred object prop) to components like `VictoryGroup` and `VictoryStack`?

```
    const descId =
      this.props["aria-describedby"] ||
      this.props.name ||  
      this.props.children.map((prop) => prop.props.name).join(", ")  || 
      desc;
```

boygirl: Oh! Actually, I really like the solution of using the `desc` string itself as a the id if one isn't given. Maybe kebab cased? Then we can remove the `container` role and avoid passing extra props.

ljones87: oh great! that makes it easy. Have made this change, just adding tests then hopefully ready for final review

boygirl:
@ljones87 This looks great! Good to merge, I think.

PR #1712, opened 2020-10-08
Feature/accessibility documentation

ljones87:
- Adds a documentation sidebar nav for `VictoryAccessibleGroup` under "More". 
- Updates primitive prop specs to include `aria-label` string/function and `tabIndex` number/function
- Add update about VictoryAccessibleGroup to "New Features" and add small example for primitive prop updates

#### `VictoryAccessibleGroup` new section
Screen Shot 2020-10-08 at 1 28 18 PM

#### `VictoryPrimitives` doc update
Screen Shot 2020-10-08 at 4 04 26 PM

### DIFF
--- a/docs/src/content/docs/victory-primitives.md
+++ b/docs/src/content/docs/victory-primitives.md
@@ -19,7 +19,7 @@ Each of these primitive components renders SVG elements. The following component
 Used by `Background`, `VictoryClipContainer`, and `Voronoi`
 
 ```jsx
-const Circle = props => <circle vectorEffect="non-scaling-stroke" {...props} />;
+const Circle = (props) => <circle vectorEffect="non-scaling-stroke" {...props} />;
 ```
 
 ### ClipPath
@@ -27,7 +27,7 @@ const Circle = props => <circle vectorEffect="non-scaling-stroke" {...props} />;
 Used by `VictoryClipContainer` and `Voronoi`
 
 ```jsx
-const ClipPath = props => (
+const ClipPath = (props) => (
   <defs>
     <clipPath id={props.clipId}>{props.children}</clipPath>
   </defs>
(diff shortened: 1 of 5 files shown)


docs/src/content/docs/victory-primitives.md
```diff
@@ -92,6 +92,7 @@ const TSpan = props => <tspan {...props} />;
 **Props**
 
 - `active` _boolean_ a flag signifying whether the component is active
+- `aria-label` _string or function_ label that will be applied to the rendered path. When this prop is given as a function it will be called with the rest of the props supplied to `Arc`
```
boygirl, line 95: These should all be `ariaLabel` instead of `aria-label`, but maybe the first line here should read "a prop that controls the aria-label that will be applied to..."

ljones87: great thank you, left over from the VictoryAccessibleGroup props
have changed all to:
 -`ariaLabel` _string or function_ aria-label that will be applied to the rendered...`

boygirl:
Looking good except for a minor naming issue. For your addition to `new.md` a more general blurb about accessibility improvements including the new `ariaLabel` prop and `VictoryAccessibleGroup`. This doesn't need to be long at all, and can just link to other sections of the docs for more detail.

ljones87:
ok cool, I had added this small bar chart to demo the primitive props, but `ariaLabel` wont apply..not sure if I need to specify that prop in the file somewhere so it is recognized? Or happy to remove it all together if easiest. 
Screen Shot 2020-10-08 at 3 28 14 PM

boygirl:
@ljones87 to get the new props to apply, you'll need to update the version of victory the docs site is using to the newly published `35.2.0`.

boygirl:
Looking great! Approved pending a couple small copy changes

boygirl:
@ljones87 Everything looks great on staging. I'm going to merge now, and this will be automatically deployed

Issue #1549, opened 2020-04-30
Static member defaultEvents on VictoryTooltip is not in TypeScript typings


Zeddox:
Following the [Custom Tooltip Labels](https://formidable.com/open-source/victory/gallery/custom-tooltip-labels) example in a TypeScript project leads to a compiler error.

The [type definition file for VictoryTooltip](https://github.com/FormidableLabs/victory/blob/master/packages/victory-tooltip/src/index.d.ts) doesn't export with this static member included.

beccanelson:
Thanks for reporting this!

The type of this `CustomLabel` component is just a `React.Component` class, which in Typescript doesn't allow static properties to be set after initialization.

We can fix this by moving the static property inside the class like this.

```
class CustomLabel extends React.Component {
  static defaultEvents = VictoryTooltip.defaultEvents;

  render() {
    return (
      <g>
        <VictoryLabel {...this.props} />
        <VictoryTooltip
          {...this.props}
          x={200}
          y={250}
          orientation="top"
          pointerLength={0}
          cornerRadius={50}
          flyoutWidth={100}
          flyoutHeight={100}
          flyoutStyle={{ fill: "black" }}
        />
      </g>
    );
  }
}
```

Let me know if this helps - we can also update the documentation to use this syntax instead.

PR #1751, opened 2020-12-01
Add static defaultEvents to VictoryTooltip type

beccanelson:
Fixes https://github.com/FormidableLabs/victory/issues/1549


boygirl (approved):
Looks good! Thanks for the fix.

PR #1756, opened 2020-12-07
Update clipContainerComponent example.

athomann:
Update clipContainerComponent example by removing unnecessary bracket.


boygirl (approved):
Thank you!

PR #1766, opened 2021-01-05
Allow bin prop to be an array of negative numbers as well

NgoKnows:
This PR updates the `bins` prop of `VictoryHistogram` to allow negative numbers as well. For example, `bins={[-20, 0, 20]}` should be allowed.


boygirl (approved):
Thanks @NgoKnows! I'll get this merged and release a patch for you tonight.

PR #1820, opened 2021-04-12
Migrate to Github actions

jhumbug:
- Migrates from Travis.ci to Github Actions
- Runs `nps check.ci` on every PR push
- Runs `chromatic` on every PR push with specific changes and the "ready" label


.github/workflows/chromatic.noyml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  chromatic:
+    name: Run Chromatic Tests
+    if: ${{ github.event.label.name == 'ready' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '15'
```
boygirl, line 18: Please use an even numbered node version here

.github/workflows/chromatic.noyml
```diff
@@ -0,0 +1,41 @@
+name: Chromatic
```
boygirl, line 1: This has an odd extension. should it be `.yml`?

boygirl:
@jhumbug this is looking good, but I'd like to see all the checks run. Can you make some example changes?

.github/workflows/chromatic.noyml
jhumbug: Yes, it should.  I disabled it temporarily for testing.  For some reason the CI actions won't run at all on this PR anymore when I push.  I'm not sure what's going on.  I'm trying a few things like disabling this one workflow.

ryan-roemer:
Are we going to add a CI check that puts CI into pending / failed state if a `ready` label is not added? We could also do something slightly more sophisticated like have a `no-e2e` or something label that puts CI in green without running chromatic...

jhumbug:
> Are we going to add a CI check that puts CI into pending / failed state if a `ready` label is not added? We could also do something slightly more sophisticated like have a `no-e2e` or something label that puts CI in green without running chromatic...

Those are good suggestions, I can look into those.  However right now I'm having a weird issue where the workflows will not run at all on this PR no matter what I push to it.  I even closed and reopened it with no effect.  I haven't seen this before, have you?  I could try closing this and just restarting a new PR altogether.

boygirl:
I don't think any CI will run on a PR with merge conflicts.

jhumbug:
> I don't think any CI will run on a PR with merge conflicts.

Yep, that was it, I didn't know that.

boygirl:
And definitely agree with @ryan-roemer. It would be great to be able to skip chromatic for some changes, especially for external contributors who won't be able to run it without access to the app code

boygirl:
@jhumbug It looks like the chromatic task is failing due to a missing token. Also, wow that check is verbose. In travis, chromatic would add custom checks like so: 

Screen Shot 2021-04-13 at 8 32 35 AM
(from https://github.com/FormidableLabs/victory/pull/1775)

It looks like chromatic has some reccommended GH workflows and have published an action we might want to use: https://www.chromatic.com/docs/github-actions

.github/workflows/ci.yml
```diff
@@ -0,0 +1,58 @@
...
+
+jobs:
+  test:
+    name: Check and build codebase
+    runs-on: ubuntu-latest
+    env:
+      DISPLAY: :99.0
+    strategy:
+      matrix:
+        node-version: [12.x, 14.x, 15.x]
```
boygirl, line 28: Please remove `15.x` since these check each take 5+ minutes

boygirl:
@jhumbug this is looking great! I want to play with it a bit more to see how it behaves for 1) external PRs that don't have access to the chromatic app code, and 2) PRs without the ready label

PR #1835, opened 2021-05-06
Axis Prop Updates

jhumbug:
This provides a fix for when the domain of a dataset using VictoryZoom changes, such as switching out data on demand.


jhumbug:
@boygirl Seems like this was a very simple fix.  Suspiciously simple.  But I couldn't find any unintended consequences of fixing it like this.  Any insights on your end?

jhumbug:
I looked into what you mentioned about offsetX and offsetY on Chart and Axis, but that didn't seem to be related as far as I could tell.

boygirl:
@jhumbug apologies! I think this does fix a portion of the issue, but I should have been clearer about the axis offset issue. Here's a gif that I think illustrates the problem:

axis-offset-issue

In the chart above, there are not user provided `xOffset` / `yOffset` props. The offsets are being calculated by `VictoryChart` based on the domain. However, when zooming, these offsets aren't changing, because `VictoryZoomContainer` alters the domain prop for all the _children_ of `VictoryChart`, but not for `VictoryChart` itself.

jhumbug:
Ok, yeh I missed that.  I will look into that part of it now, thank you for the gif.

packages/victory-axis/src/helper-methods.js
```diff
@@ -296,12 +372,28 @@ const getCalculatedValues = (props) => {
...
+  const ticks = Axis.getTicks(props, currentAxisScale, props.crossAxis);
+  const tickFormat = Axis.getTickFormat(props, currentAxisScale);
   const anchors = getAnchors(orientation, isVertical);
 
+  const range = {
+    x: Helpers.getRange(props, "x"),
+    y: Helpers.getRange(props, "y")
+  };
+  const domain = {
+    x: Domain.getDomain(props, "x"),
```
boygirl, line 385: I'm not sure `Domain.getDomain` is going to do the same thing as `Axis.getDomain`, since the generic `getDomain` doesn't care about `tickValues`.  I think `Axis.getDomain` takes an optional `axis` prop (i.e. x or y), so I think you can drop in replace that here.

packages/victory-axis/src/helper-methods.js
```diff
@@ -169,42 +169,116 @@ const getLabelPadding = (props, style) => {
...
-    const tick = stringTicks ? props.tickValues[data - 1] : data;
-    const tickStyle = Helpers.evaluateStyle(style.ticks, assign({}, sharedProps, { tick, index }));
-    return tickStyle.size || 0;
-  });
-  const totalPadding = fontSize + 2 * Math.max(...tickSizes) + labelPadding;
-  const minimumPadding = 1.2 * fontSize; // eslint-disable-line no-magic-numbers
-  const x = isVertical ? totalPadding : minimumPadding;
-  const y = isVertical ? minimumPadding : totalPadding;
+  const { top, bottom, left, right } = padding;
+  const origin = polar ? Helpers.getPolarOrigin(props) : Axis.getOrigin(domain);
```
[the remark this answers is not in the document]
boygirl: Also, I think `Helpers.getPolarOrigin` can be removed, since there is a totally separate `VictoryPolarAxis`. This component should never be polar.

packages/victory-chart/src/helper-methods.js
```diff
@@ -17,9 +17,6 @@ function getAxisProps(child, props, calculatedProps) {
   const axis = child.type.getAxis(childProps);
   const crossAxis = childProps.crossAxis === false ? false : true;
```
boygirl, line 18: Since we're calculating the full domain in `VictoryAxis` I think we can also easily let it handle `crossAxis` and `orientation` by itself now.

packages/victory-axis/src/helper-methods.js
boygirl: Thinking about this more, I think the correct behavior is actually something like:

```
domain = {
  x:  props.domain && props.domain[x] || (currentAxis === "x" && Axis.getDomain(props) || undefined,
  y: ...
}
```
So that it will 1) use the domain given to it from its parent if it exists, 2) calculate its own domain (only for its correct dimension if it isn't given a parent domain, and 3) _not_ try to calculate a domain for the opposite dimension if it's not being used within a chart because in that case it can't know anything about the opposite axis  (because it won't have access to a shared domain) and shouldn't try.

boygirl:
From looking at the stories, it looks like there's a systematic problem with horizontal charts and with how the range / padding is being calculated

jhumbug:
> From looking at the stories, it looks like there's a systematic problem with horizontal charts and with how the range / padding is being calculated

Yep, addressing those now. Most are the horizontal issue which is easy.  I'll look at range/padding after.

jhumbug:
Got most of the chromatic issues with those fixes.  I'll work on the few left tomorrow.

jhumbug:
The last chromatic changes are from added/changed tests in previous PRs and 2 (VictoryScatter: Theme & VictoryLine: Theme) that actually look correct now, rather than the previous version.  They have slight differences in the grid position, but it looks like the new versions are right now, no?

packages/victory-axis/src/helper-methods.js
```diff
@@ -257,21 +340,28 @@ const getGridEdge = (props, calculatedValues) => {
...
-    y: props.crossAxis ? offset.y - yPadding : 0
+    x: crossAxis ? offset.x - xPadding : 0,
+    y: crossAxis ? offset.y - yPadding : 0
   };
 };
 
 const getLayoutProps = (modifiedProps, calculatedValues) => {
-  const offset = getOffset(modifiedProps, calculatedValues);
+  let offset;
+  if (modifiedProps.standalone) {
```
boygirl, line 355: Are you using `standalone` as a check for whether the axis is a child of `VictoryChart`? That's not a perfect check for that, unfortunately. Maybe it's worth adding a flag for that, but for now, a better check would be the existence of x and y domain values.

packages/victory-chart/src/helper-methods.js
```diff
@@ -224,73 +200,18 @@ const getDomain = (props, axis, childComponents) => {
...
-  return {
-    x: originPosition.x ? Math.abs(originOffset.x - originPosition.x) : orientationOffset.x,
-    y: originPosition.y ? Math.abs(originOffset.y - originPosition.y) : orientationOffset.y
-  };
-};
-
-const getHorizontalAxisOffset = (props, calculatedProps, orientation) => {
-  const { scale, origin, domain, padding } = calculatedProps;
-  const { top, bottom, left, right } = padding;
+function getOrientation(axis, originSign, horizontal) {
```
boygirl, line 203: Are orientations still being used for anything in VictoryChart?

boygirl:
@jhumbug A couple minor comments, but this is otherwise looking ready. And axes look _so much nicer_ when zooming

better-zooming

packages/victory-axis/src/helper-methods.js
jhumbug: I was, yes.  Wasn't 100% sure of that, good to know. I'll switch it to domain.

packages/victory-chart/src/helper-methods.js
jhumbug: I wasn't sure if another type of graph that VictoryChart might wrap used orientation so I left it in.  I searched globally in the project and saw some things that might, but I'd have to really trace through to be sure. I'll look again.

packages/victory-axis/src/helper-methods.js
jhumbug: done

packages/victory-chart/src/helper-methods.js
jhumbug: Looks like nothing does, I removed it.

PR #1880, opened 2021-06-15
Disable inline styles

beccanelson:
This is the work to add a `disableInlineStyles` prop to all chart components in order to work better with CSS styles and styled-components. 

Some components have required more customization, but a lot of components use the shared `Helpers.evaluateStyle` function, so changing this function did a lot of the work for us.


docs/src/content/common-props/common-props.md
```diff
@@ -165,9 +165,9 @@ _default:_ `containerComponent={<VictoryContainer/>}`
 
 Specify data via the `data` prop. By default, Victory components expect data as an array of objects with `x` and `y` properties. Use the [x][] and [y][] data accessor props to define a custom data format. The `data` prop must be given as an array. Data objects may also include information about ~~styles~~, labels, and props that may be applied to individual data components.
 
-*note:* All values stored on the data object will be interpolated during animation. Do not store functions on data objects.
+_note:_ All values stored on the data object will be interpolated during animation. Do not store functions on data objects.
```
beccanelson, line 168: Apparently prettier likes this formatting better now? 🤷‍♀️

docs/src/content/common-props/common-props.md
```diff
@@ -226,6 +226,12 @@ class App extends React.Component {
 ReactDOM.render(<App/>, mountNode);
 ```
 
+## disableInlineStyles
+
+`type: boolean`
+
+The `disableInlineStyles` prop allows Victory components to work better with CSS classes or styled-components. By default, Victory provides inline styles to chart components, which will override any conflicting CSS classes. This flag will remove the inline styles, making it easier to provide custom styling for components.
+
```
beccanelson, line 234: Since this prop is needed for multiple components, I added it to the common prop types. An example here might be helpful as well.

stories/victory-line.stories.js
```diff
@@ -455,3 +456,22 @@ export const Polar = () => {
...
+    <div style={containerStyle}>
+      <VictoryChart>
+        <VictoryLine disableInlineStyles />
+      </VictoryChart>
+      <VictoryChart>
+        <VictoryLine dataComponent={<StyledCurve disableInlineStyles />} />
+      </VictoryChart>
+    </div>
+  );
+};
```
beccanelson, line 477: This storybook component isn't actually working yet because it pulls the `evaluateStyle` function from the helpers in the latest version of`victory-core`, which won't actually be updated until this PR is merged. I'm not sure if there's a better way to test these changes for now?

boygirl: @beccanelson if you start a dev server (`yarn start`) and leave it running at the same time that you have your storybook open, all of your packages will rebuild on changes, and all the package linking for local changes will be taken care of by lerna. 

Looking more like you would expect? 
Screen Shot 2021-06-16 at 10 38 26 AM

boygirl:
@beccanelson can you alter your PR so that it is against the feature/disable-inline-styles branch rather than main so it's easier to see what needs review and to prevent building docs until we're all done.

stories/victory-line.stories.js
beccanelson: Yes, thank you!

beccanelson:
Closing in favor of #1882

PR #1912, opened 2021-07-23
add a check for nested children before attempting barWidth calculation

WaysToGo:


boygirl (approved):
Thanks! I'll get another patch out for you today

PR #1922, opened 2021-08-02
Proof of Concept: Replace addEvents HOC with useEvents hook

beccanelson:
This is an experimental branch for replacing the addEvents HOC with a useEvents hook. In its current form, this cuts down on manual caching and lifecycle methods, but does not include full functionality yet. So far there is a pretty dramatic performance improvement with large datasets (~5000 points), but more code will be required to reach full functionality.


beccanelson:
Closing this for now, but I might come back to it later!

PR #1931, opened 2021-08-11
VictoryGroup refactoring + performance improvements

### LINKED ISSUE #1927: Update VictoryGroup so it is not re-calculating props each time its children change
Right now, `VictoryGroup` is rerendering and recalculating its props each time it's children change, which adds around 200ms to each render cycle when there is some interaction. If I change the return value of `shouldComponentUpdate` to false after the initial render, this cuts out this render cycle (while still updating the child components). We can either update `shouldComponentUpdate` with a different conditional, or refactor this component to use memoization. 

Screen Shot 2021-08-05 at 6 24 07 PM

beccanelson:
This will need more QA before it is ready to merge, but there is a noticeable improvement here in re-rendering performance. We might have been able to achieve the same effect by updating `shouldComponentUpdate`, but I wanted to refactor to a function so we can use `useMemo` and `useCallback` in other places as well.

Previously, each re-render cycle (in development mode) was taking around 225ms. This refactor cuts each cycle down to ~9.5ms.

Before:

Screen Shot 2021-08-11 at 2 00 41 PM

After:

Screen Shot 2021-08-11 at 1 59 20 PM

Closes #1927


beccanelson:
Looks like there is some interesting stuff happening on Chromatic! I will look into it.

packages/victory-group/src/victory-group.js
```diff
@@ -19,150 +21,209 @@ const fallbackProps = {
...
-  static getChildren = getChildren;
-
-  constructor(props) {
-    super(props);
-    if (props.animate) {
-      this.state = {
-        nodesShouldLoad: false,
-        nodesDoneLoad: false,
-        animating: true
+const usePreviousProps = (props) => {
```
beccanelson, line 24: This might be a more generic helper that we can put somewhere else

packages/victory-group/src/victory-group.js
```diff
@@ -19,150 +21,209 @@ const fallbackProps = {
...
+    horizontal,
+    name
+  } = modifiedProps;
+
+  const childComponents = React.Children.toArray(modifiedProps.children);
+  const calculatedProps = getCalculatedProps(modifiedProps, childComponents);
+  const { domain, scale, style, origin } = calculatedProps;
+
+  // This is a copy of Wrapper.getAnimationProps that doesn't use this.setState
+  const getAnimationProps = React.useCallback(
```
beccanelson, line 56: We might want to look into refactoring the `Wrapper.getAnimationProps` function so it doesn't call setState inside a getter function, but it would be helpful to have a test case so I know I'm not breaking anything!

packages/victory-group/src/victory-group.js
```diff
@@ -19,150 +21,209 @@ const fallbackProps = {
...
+
+VictoryGroup.expectedComponents = [
+  "groupComponent",
+  "containerComponent",
+  "labelComponent"
+];
+
+VictoryGroup.getChildren = getChildren;
+
+export default React.memo(VictoryGroup, isEqual);
```
boygirl, line 229: I'm curious about whether this memoization is worth the potentially expensive equality check

beccanelson: I had the same question! Right now, it seems to be saving us about 200ms of render time on each interaction—if we take it out the performance is basically on par with the initial measurements. If we're only calling isEqual once per VictoryGroup component, it seems to have a pretty miminal performance impact. The profiler says it takes around 0.1 ms per function call.

Screen Shot 2021-08-11 at 2 39 44 PM

packages/victory-group/src/victory-group.js
boygirl: I don't _think_ the `setState` is actually called from within this getter function. I think it's just used by `Transitions.getTransitionPropsFactory` to set up the functions (including appropriate `setState`) that will be called for each transition. 

Either way, this refactor looks cleaner to me, and I haven't been able to identify anything breaking from the change. If we're moving to using hooks for this, though, it's probably worth also looking at `Wrapper.setAnimationState`

boygirl:
From the chromatic tests, I noticed that this refactor isn't including the `getDefaultDomainPadding` calculation that is included in `Wrapper`, here: https://github.com/FormidableLabs/victory/blob/main/packages/victory-core/src/victory-util/wrapper.js#L34

Screen Shot 2021-08-11 at 2 51 57 PM

packages/victory-group/src/victory-group.js
boygirl: Nice!

beccanelson:
@boygirl After a bunch of debugging, it turns out that VictoryChart was incorrectly calculating the children's domain, since it couldn't determine the child's static role property when it was wrapped in `React.memo`. I switched some things around, and we will see if that fixes Chromatic.

boygirl:
@beccanelson This is all looking good, and I think it's ready to merge. I'd love to see this replicated in `VictoryStack` and `VictoryChart` before releasing it as a minor version. Would you prefer a feature branch instead of `main` for that?

beccanelson:
@boygirl Sure, let's create a feature branch for this larger refactoring effort for these three components. That way you can continue to release other improvements on main I can solidify some of these patterns before merging to main.

PR #1937, opened 2021-08-16
Refactor VictoryStack

beccanelson:
This applies the same refactor from VictoryGroup in #1931 to VictoryStack. This one has about the same before and after performance, because we weren't seeing as much re-rendering with VictoryStack originally.

Before: 

Screen Shot 2021-08-16 at 3 01 00 PM

After:

Screen Shot 2021-08-16 at 2 57 35 PM


boygirl (approved):
Looks great, and everything is working well locally. Thank you!

PR #1940, opened 2021-08-17
Refactor and memoize parent components (VictoryGroup, VictoryStack, VictoryChart)

### LINKED ISSUE #1927: Update VictoryGroup so it is not re-calculating props each time its children change

beccanelson:
This PR converts all three of these components to functions so we can use React.memo to prevent them from re-rendering and recalculating values unnecessarily.

VictoryStack and VictoryContainer are mostly on par with the original components in terms of performance, but VictoryGroup includes a few performance improvements that I detailed in #1931. 

Fixes #1927

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,11 @@ module.exports = {
       }
     }
   },
-  extends: ["formidable/configurations/es6-react", "prettier"],
+  extends: [
+    "formidable/configurations/es6-react",
+    "prettier",
+    "plugin:react-hooks/recommended"
+  ],
   rules: {
     "react/sort-comp": "off",
     "import/no-unresolved": [2, { ignore: ["victory*"] }],
--- a/demo/js/components/animation-demo.js
+++ b/demo/js/components/animation-demo.js
@@ -110,19 +110,19 @@ export default class App extends React.Component {
 
     return (
       <div className="demo" style={containerStyle}>
-        <VictoryArea
+        {/* <VictoryArea
           style={style}
           animate
           data={this.state.areaTransitionData}
           x={(d) => d.x}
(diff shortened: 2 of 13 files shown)


boygirl:
@beccanelson Thanks for this work! This is mostly looking correct, but I've just noticed a regression with exit transitions for wrapped components. Here are a couple gifs from the animation demo page

on `main`:
exit-animations-on-main

and on this branch:
exit-animation-regression

This is honestly one of the most annoying parts of animating transitions, and might be a bit obnoxious to track down. Basically, though, when nodes are exiting, we need to update the `domain` based on the new props, but keep `data` from the old props until the transition finishes.

beccanelson:
@boygirl Hmm, my guess right now is that this might be due to the slightly different logic between `shouldComponentUpdate` (which is called before a re-render) and `useEffect` (which is called right after a re-render). I will investigate more!

beccanelson:
@boygirl I fixed the domain issue (and discovered a couple other bugs in the process)! We needed to call `setAnimationState` as a return function in useEffect rather than just calling it as an effect. This will call it before the component dismounts, which is similar to `shouldComponentUpdate`. Before, the domain wasn't getting updated until after render, which wasn't in sync with the previous state change.

beccanelson:
Merging!

PR #1943, opened 2021-08-24
Update wrapper export syntax

beccanelson:
Removes default export in wrapper.js and removes unused functions (that have been moved to hooks.js). It's a bit hard to read the diff, but the functions have just been re-ordered in order to prevent calling the function before declaring it.

This also removes the unused sinon spy that was causing an error (there were no assertions on the sandbox).


boygirl (approved):
Thanks for this clean up!

PR #1961, opened 2021-09-03
Refactor and memoize parent components (VictoryGroup, VictoryStack, VictoryChart)

beccanelson:
Reverts FormidableLabs/victory#1949

This re-instates the previous refactor, and this time we will consider it a major version bump. Before we bump the version on victory native, we will need to update any components that are extending these three components, as they are no longer classes. 

Original PRs: #1940, #1943


boygirl (approved):

boygirl:
@beccanelson thank you! I'll get this released as a _major_ version this time :)

PR #1971, opened 2021-09-29
Add canvas line primitives

beccanelson:
This adds a `CanvasContainer` and `Curve` component that allows the existing `VictoryLine` component to render using the canvas API. 

So far, this includes a basic demo and some duplicated stories from VictoryLine. 

To test, run `yarn storybook` and navigate to Victory Canvas/Line. The stories should be nearly identical to the existing VictoryLine stories.


beccanelson:
I'm going to merge these smaller PRs for now, and we can do a more extensive review of the feature branch once there is a v1 ready to merge.

demo/js/components/canvas-demo.js
```diff
@@ -0,0 +1,263 @@
...
+          {populationData.map(({ country, values }) => {
+            const data = values.map(({ year, value }) => ({
+              x: year,
+              y: value
+            }));
+            return (
+              <VictoryLine
+                key={country}
+                data={data}
+                groupComponent={<CanvasContainer />}
```
boygirl, line 250: I really like providing the canvas as the `groupComponent` here so that we leave open the possibility of using the interactive containers with canvas. This might be a motivation for coming up with a different name than `CanvasContainer` though, to avoid confusion about how it should be used.

packages/victory-canvas/src/curve.js
```diff
@@ -0,0 +1,30 @@
...
+
+const Curve = (props) => {
+  const { canvasRef, clear } = useCanvasContext();
+  const { style, data } = props;
+  const { stroke, strokeWidth } = style;
+
+  const draw = React.useCallback(
+    (ctx) => {
+      const line = LineHelpers.getLineFunction(props);
+      ctx.strokeStyle = stroke;
```
boygirl, line 13: It might be worth thinking about a generalized svg to ctx style mapping helper

packages/victory-canvas/src/curve.js
```diff
@@ -0,0 +1,30 @@
...
+      ctx.lineWidth = strokeWidth;
+      ctx.beginPath();
+      line.context(ctx)(data);
```
boygirl, line 16: Really wonderful that the d3 shape functions just work

beccanelson: Yes!

demo/js/components/canvas-demo.js
beccanelson: Yes, that's a good thought!

PR #1972, opened 2021-10-01
Adds canvas scatter and bar charts

beccanelson:
This implements basic scatter and bar charts using canvas.

As of right now, these primitive components do not include:
+ Animations (I'm working on that)
+ Scatter point shapes other than circles
+ Polar bar charts (I think polar line charts work)

### DIFF
--- a/packages/victory-canvas/src/point.js
+++ b/packages/victory-canvas/src/point.js
@@ -1,0 +1,73 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { assign } from "lodash";
+import { Helpers, CommonProps } from "victory-core";
+import { useCanvasContext } from "./hooks/use-canvas-context";
+
+const evaluateProps = (props) => {
+  /**
+   * Potential evaluated props are:
+   * `ariaLabel`
+   * `desc`
+   * `id`
+   * `size`
+   * `style`
+   * `symbol`
+   * `tabIndex`
+   */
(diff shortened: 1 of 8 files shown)


packages/victory-canvas/src/point.js
```diff
@@ -0,0 +1,73 @@
...
+const Point = (initialProps) => {
+  const { canvasRef } = useCanvasContext();
+  const props = evaluateProps(initialProps);
+  const { x, y, style, size } = props;
+
+  const draw = React.useCallback(
+    (ctx) => {
+      ctx.fillStyle = style.fill;
+      ctx.beginPath();
+      ctx.arc(x, y, size, 0, 2 * Math.PI);
```
boygirl, line 46: It looks like this is confined to circle points only so far. It would be cool to eventually be ablr to translate from arbitrary svg paths to canvas commands. Basically the reverse of what `d3-path` does: https://github.com/d3/d3-path/blob/main/src/path.js

beccanelson: Yes! I have been looking into using something like d3-shape for this. It was pretty cool how we could use the exact same logic for the lines and just set the context to a canvas container instead of SVG, and I'm not sure yet whether there is a way to do this for other shapes.

beccanelson:
Merging to the feature branch for now, and making note of the feedback to include in my next PR.

PR #1974, opened 2021-10-01
Initial Prototype of Native Repo Absorption

gksander:
## Sorry -- this thing is _huge_

This is the initial work of moving `victory-native` into the `victory` monorepo, with an associated demo.

### What's included in this PR:

- `victory-native` was added to the `packages`.
- `demo/rn` was added. An Expo demo app that consumes the development version of `victory-native`.
- `victory-native` was patched to get it working with the latest changes to `victory` core packages.

### What's left to do:

Plenty. Here's a non-exhaustive list:

- For the time being, exclude `victory-native` from the lerna publishing pipeline, until we're 100% ready to publish `victory-native` through here.
- Add tests to `victory-native` and add that test into the CI pipeline. This might even be as simple as using `jest` and `@testing-library/react-native` to just mount each component to ensure that they'll mount (and then potentially expanding this test suite).
- Ensure that the move of `victory-native` into the monorepo, and the recent changes to the core packages, didn't break `victory-native` in unintended ways.
- Eventually, start sunset the original `victory-native` repo and document this process, and start publishing the native library from this lerna setup.


demo/rn/metro.config.js
```diff
@@ -0,0 +1,48 @@
+const path = require("path");
```
gksander, line 1: This file is the magic that allows for importing the development files into the demo app, even though they live two levels above the demo app.

packages/victory-native/src/helpers/wrap-core-component.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+/**
+ * Wrap a core component, pass props through.
+ * @param {any} Component Core Component
+ * @param {object} defaultProps Default props for component
+ * @returns {React.FC} WrappedComponent Wrapped component (passes props through)
+ */
+export const wrapCoreComponent = ({ Component, defaultProps }) => {
```
gksander, line 9: This is a helper method that wraps up a core component, and sets default props accordingly. This is used instead of class extension to allow the conversion of class components to functional components.

demo/rn/src/screens/veronoi-screen.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from "react";
+import { ScrollView } from "react-native";
+import viewStyles from "../styles/view-styles";
+import { VictoryTheme, VictoryVoronoi, VictoryChart } from "victory-native";
+
+export const VeronoiScreen: React.FC = () => {
```
boygirl, line 6: NIT: "voronoi" not "veronoi"

packages/victory-native/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "victory-zoom-container": "^36.0.0"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.15.5",
+    "react-native": "^0.65.1",
+    "react-native-svg": "^12.1.1",
+    "react-native-gesture-handler": "^1.10.3"
+  },
+  "scripts": {
+    "version": "yarn run nps build-libs && nps build-dists"
```
boygirl, line 58: We're probably not going to build these when publishing native. We publish unaltered src only at the moment and I imagine that will remain so

packages/victory-native/src/components/victory-container.js
```diff
@@ -0,0 +1,156 @@
...
+import { assign, get } from "lodash";
+import { View, PanResponder } from "react-native";
+import { VictoryContainer } from "victory-core/es";
+import NativeHelpers from "../helpers/native-helpers";
+import Portal from "./victory-portal/portal";
+
+const yes = () => true;
+const no = () => false;
+
+export default class extends VictoryContainer {
```
boygirl, line 13: TODO: we should investigate getting away from class extension here as well.

packages/victory-native/src/helpers/wrap-core-component.js
boygirl: I generally like this approach. It breaks down for some of what we had been doing with class extension, (i.e. the containers), but we can look into other ways to avoid that pattern for those.

boygirl:
@gksander after your walkthrough this morning, I'm comfortable merging this as the baseline for the victory-native feature branch. I think the remaining integration todos, (especially lerna workflow integrations) will inform the places that need more thorough review.

packages/victory-native/src/components/victory-container.js
gksander: Agreed! There were a few places where there's class extension, but in a non-trivial way, and tackling those will require a little more attention. I figured those should probably happen in their own unit of work to ensure the conversion/refactoring doesn't come with any regressions.

demo/rn/src/screens/veronoi-screen.tsx
gksander: 😆 good catch. I've never seen that word before, and I let my fingers take the wheel. I misspelled that multiple times (I was wondering why my editor spelling suggestions were mad about "veronoi" lol). Just patched those.

PR #1977, opened 2021-10-03
Add explicit types for VictoryGroup#children

eps1lon:
We plan to remove implicit types for `children` in `@types/react@18.x`. 

I ran `yarn nps compile-ts` and only `VictoryGroup` was reporting failures when `children` weren't added by `@types/react`.


boygirl:
@eps1lon Thanks for this improvement!

Issue #1758, opened 2020-12-15
How is possible to wrap group (and other) components? (for Data Driven VictoryCharts)


brumik:
# Question

### Checklist

- [x] This is not a `victory-native` specific issue. (Issues that only appear in `victory-native` should be opened [here](https://github.com/FormidableLabs/victory-native/issues/new))

- [x] I have read through the [FAQ](https://formidable.com/open-source/victory/docs/faq) and [Guides](https://formidable.com/open-source/victory/guides/) before asking a question

- [x] I am using the latest version of Victory

- [x] I've searched open issues to make sure I'm not opening a duplicate issue

### Description
I am trying to render charts from a huge JSON file. For this, to work I am wrapping the Victory components into other components (which handles some customizations renders the component's children via custom logic). 

The easiest I think is to show on a basic example (the `Component` element in the next examples are assigned dynamically)

It works for:
```typescript
const Component = (props) => (<VictoryBar data={ ... } {...props} />)

<VictoryChart>
    /* set the axis domain according to the data  */
    <Component />
</VictoryChart>
```
It is still imperfect since I could not figure out how the `VictoryChart` components see the `data` passed to the `VictoryBar` when not wrapped into `Component`. I would like to achieve the same functionality if possible. 

However, when I try groupped chart it displays only one data set: (offset for data sets are not applied)
```typescript
const Component = (props) => (<VictoryBar data={ ... } {...props} />);
const GroupComponent = (props) => (<><Component /><Component /></>);

<VictoryChart>
    /* set the axis domain according to the data  */
    <GroupComponent />
</VictoryChart>
```

I am struggling to understand how the implementation works, or how is the data passed around in the `Chart` or `Group` component and how could I achieve the same functionality when I am wrapping these components.

jordantomax:
Is there any update on this issue? I am having a similar issue, attempting to wrap a set of Victory components in a custom component. I am seeing unexpected behavior, even when following the [guide on wrapping components](https://formidable.com/open-source/victory/guides/custom-components#wrapping-components). My components are structured like this:

```javascript
function Outside () {
  return (
    <VictoryChart>
      <VictoryAxis scale='time' />
      {arr.map(item => <Inside item={item} />)}
    </VictoryChart>
  )
}

function Inside ({ item }) {
  const data = item.points.map(point => ({
    date: new Date(point.date),
    value: point.value
  }))

  return (
    <VictoryGroup data={data}>
      <VictoryAxis dependentAxis />
[...]

PR #1999, opened 2021-10-15
Add recursive check for the first child

cneltyn:
The https://github.com/FormidableLabs/victory/pull/1908/commits commit fixes the issue https://github.com/FormidableLabs/victory/issues/1911

But there we consider a group of bars. Actually, we can use a group of _stacked_ bars. Previously (at least with 35.4.8) it worked correctly.

### DIFF
--- a/packages/victory-core/src/victory-util/wrapper.js
+++ b/packages/victory-core/src/victory-util/wrapper.js
@@ -172,7 +172,13 @@ export function getDefaultDomainPadding(props, axis, childComponents) {
     return undefined;
   }
 
-  const firstChild = Array.isArray(children) && children[0];
+  function getFirstChild(childrenComponents) {
+    if (Array.isArray(childrenComponents)) {
+      return getFirstChild(childrenComponents[0]);
+    }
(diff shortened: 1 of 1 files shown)


beccanelson:
I'm going to close this for now, since fixing the nested array in the consumer is probably a better solution than adding complexity to the library. Let me know if there is any further discussion to be had here!

cneltyn:
> I'm going to close this for now, since fixing the nested array in the consumer is probably a better solution than adding complexity to the library. Let me know if there is any further discussion to be had here!

@becca-bailey, I believe, the workaround with array fix the issue described [here](https://github.com/FormidableLabs/victory/issues/1758). Actually, we have the same case with lots of data, that we group into stacks of bars and scatters. Previously it works for us, but now we need the newer version of the library, that fixes the domain padding [issue](https://github.com/FormidableLabs/victory/pull/1084) we tried to help to fix previously as well. Please, have a look at it more generally. 
I tried to move the groups into separate functional components but was still unsuccessful as it renders always stacks instead of bars/stacks, as I got with the array approach.

PR #2012, opened 2021-11-01
Native tests!

gksander:
This PR ports over the test from the `victory-native` repo into this one! A couple things to note:

- I decided to go with Jest as the test runner here, as I expect a potential migration from mocha/chai over to Jest.
- I added a `jest` script to the `package-scripts`, but they're tailored to the native tests. Eventually there might be more scripts pertain to the core libraries.
- There is jest/babel setup specifically for the native tests, since testing RN libraries is... hard... and requires additional setup.

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -17,6 +17,10 @@ module.exports = {
       watch: "karma start --auto-watch ./config/karma/karma.conf.js",
       default: "karma start ./config/karma/karma.conf.js"
     },
+    jest: {
+      native: "jest --config=jest-native-config.js",
+      default: "jest --config=jest-native-config.js"
+    },
     test: {
       cov: npsUtils.series.nps("build-package-libs", "karma.cov"),
       dev: "karma start ./config/karma/karma.conf.dev.js",
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -89,13 +89,15 @@ module.exports = {
     "build-es": npsUtils.series.nps("clean.es", "babel-es"),
     "build-lib": npsUtils.series.nps("clean.lib", "babel-lib"),
     "build-libs": npsUtils.series.nps("build-lib", "build-es"),
-    "build-package-libs": "lerna exec --parallel -- nps build-libs",
+    "build-package-libs":
+      "lerna exec --parallel --ignore victory-native -- nps build-libs",
     "build-dist-dev":
       "webpack --bail --config ../../config/webpack/webpack.config.dev.js",
     "build-dist-min":
       "webpack --bail --config ../../config/webpack/webpack.config.js",
     "build-dists": npsUtils.concurrent.nps("build-dist-min", "build-dist-dev"),
     "build-dist": npsUtils.series.nps("clean.dist", "build-dists"),
-    "build-package-dists": "lerna exec --parallel -- nps build-dists"
+    "build-package-dists":
+      "lerna exec --parallel --ignore victory-native -- nps build-dists"
   }
 };
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -60,6 +60,7 @@ module.exports = {
         "build-package-libs",
         "build-package-dists",
         "karma.ci",
+        "jest",
         "compile-ts"
       ),
       cov: npsUtils.series.nps("lint", "test.cov"),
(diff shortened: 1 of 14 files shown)


package-scripts.js
```diff
@@ -17,6 +17,10 @@ module.exports = {
       watch: "karma start --auto-watch ./config/karma/karma.conf.js",
       default: "karma start ./config/karma/karma.conf.js"
     },
+    jest: {
+      native: "jest --config=jest-native-config.js",
+      default: "jest --config=jest-native-config.js"
+    },
```
gksander, line 23: I'm not 100% sure what this should look like, so I left it like this until we visit the Lerna setup more generally.

boygirl:
@gksander running `nps jest` in this branch gets the tests to run, but most of them are failing with errors like:
 ``` 
Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```
Do I need to perform a step before attempting to run these tests?

**Comment by boygirl** [comment]
hmm... I'm actually getting errors on `yarn start` in this branch as well

**Comment by ryan-roemer** [comment]
I'm wondering why CI isn't running?

**Comment by ryan-roemer** [comment]
Oh! I figured it out.

@gksander can you add `- feature/absorb-native-repo` below `- main` in https://github.com/FormidableLabs/victory/blob/main/.github/workflows/ci.yml#L15 with a `# TODO: REMOVE when <INSERT REASON>` comment so we can get CI on the long running branch? Thanks!

**Comment by gksander** [comment]
Weird, I'm going to nuke everything on my end and see if I can reproduce this.

**Comment by gksander** [comment]
Lauren, I resolved the 

```
Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

issue you were having (I must have lucked out with some cache or something, once I cleaned out some directories it started showing up).

Now I'm working on patching the `server.test` setup to stop throwing errors relative to VN.

**Comment by gksander** [comment]
@boygirl and @ryan-roemer - thanks so much for checking this out, and catching some (a lot of...) issues! I tweaked scripts to exclude certain VN things where appropriate, and added the jest script into the `check.ci` script. Now CI is running, and appears to be passing. 

Victory Native is excluded from the build scripts - which is what we intended anyways, and clears up a lot of the errors that were being thrown.

File: test/jest/components.test.js [hunk 2]
```diff
@@ -0,0 +1,55 @@
...
+  { component: VictoryStack, name: "VictoryStack" },
+  { component: VictoryTooltip, name: "VictoryTooltip" },
+  { component: VictoryVoronoi, name: "VictoryVoronoi" }
+];
+
+describe("Default render", () => {
+  components.forEach((C) => {
+    it(`should work for ${C.name}`, () => {
+      const wrapper = enzyme.shallow(React.createElement(C.component));
+      expect(wrapper).toHaveLength(1);
```
beccanelson, line 52: For new tests, is there a particular reason to use enzyme over React testing library for these kinds of rendering checks?

beccanelson, line 52: For newer tests I'm also curious whether it's worth swapping enzyme for React testing library. Not an urgent concern right now, but it does seem to be the library more people are familiar with these days.

beccanelson:
Once I updated my node version, this worked for me!

beccanelson (approved):

ryan-roemer (approved):
🎉

test/jest/components.test.js
gksander: I totally agree! I made a very brief attempt to get RN testing lib working, but ended up battling some things -- so I decided just to roll the old tests over, and try again with RNTL after this initial work is done.

The upshot: these enzyme tests, at the very least, will catch very egregious errors that would make components not mount at all.

beccanelson: Sounds good! I agree that these tests still have value, and there's no need to do the extra work to get everything working right now.

boygirl (approved):
It's working perfectly with the latest changes

PR #2013, opened 2021-11-03
Absorb victory-native README/docs

gksander:
This PR moves the Victory Native README into the `packages/victory-native` package here, and updates some of the demo instructions accordingly.

### DIFF
--- a/packages/victory-native/README.md
+++ b/packages/victory-native/README.md
@@ -1,0 +1,98 @@
+# Victory Native
+
+[![Maintenance Status][maintenance-image]](#maintenance-status)
+
+### Usage
+
+Install victory-native:
+```sh
+$ npm install victory-native --save
+```
+
+Install react-native-svg:
+```sh
+$ npm install react-native-svg --save
+```
+
+Link react-native:
+```sh
+$ react-native link react-native-svg
+```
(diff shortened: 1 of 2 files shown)


packages/victory-native/README.md
```diff
@@ -0,0 +1,98 @@
...
+# Open up the React Native demo app
+$ cd demo/rn
+$ yarn install
+$ yarn start
+
+# Run the demo from a new terminal window
+$ cd demo/rn && npx react-native run-ios
+```
+
+Changes to the Victory and Victory Native source code will be reflected in the demo app.
```
gksander, line 78: This is effectively the only thing I changed in this file, which just boils down to changing demo location and install instructions.

ryan-roemer: MINOR: Is `react-native` otherwise available in deps already from another path? A lot of the times `npx` can add a lot of unneeded install time for big packages that are already installed nearby....

gksander: Oh hey, that's a great point! And since we moved from Vanilla RN to Expo, this command is outdated anyways. Updating now!

boygirl (approved):
Thanks!

Issue #2042, opened 2021-12-23
A full zoom-out on combned Brush and Zoom chart results in incorrect brush selection being drawn


Chris7:
# Bugs and Questions

### Checklist

- [x] I have read through the [FAQ](https://formidable.com/open-source/victory/docs/faq) and [Guides](https://formidable.com/open-source/victory/guides/) before asking a question

- [x] I am using the latest version of Victory

- [x] I've searched open issues to make sure I'm not opening a duplicate issue

### The Problem

Using the brush-and-zoom chart, when performing a full zoom-out following a selection via the brush, the brush component will "overflow" and then get reset to the initial cached selection

### Reproduction

You can actually use your example to reproduce this: https://formidable.com/open-source/victory/guides/brush-and-zoom/

1) Select an area using the brush
2) Do a full zoomout using the mouse wheel
3) The zoom chart will correctly render, but the brush selection will be incorrect

![image](https://user-images.githubusercontent.com/963909/147287304-e70cb92e-a693-4774-aee0-c3b31a80ed99.png)

becca-bailey:
Interesting! I'm not actually able to reproduce this in Chrome or Safari. What browser are you using?

Chris7:
Hi Becca,

Thanks for the quick response! I'm using Firefox on Linux. I can try and
see if it reproduces in chrome as well and maybe see how to make a gif of
what I see.

Chris7:
Here's a gif reproducing the bug on the demo page:

![victory bug](https://user-images.githubusercontent.com/963909/147686078-6030d327-a3a8-4e5d-a70a-fa1b51ae5eed.gif)

becca-bailey:
Closing in favor of #2111

PR #2044, opened 2021-12-23
Always use brushDomain to avoid stale data

Chris7:
This addresses #2042 

The issue I observed was `brushDomain` and `cachedBrushDomain` are equal (and correct in their values), but `currentDomain` is set to an incorrect value. I don't know enough to know how to fix this further upstream (or how hard it would be to track how all the different state components are being set).

### DIFF
--- a/packages/victory-brush-container/src/brush-helpers.js
+++ b/packages/victory-brush-container/src/brush-helpers.js
@@ -257,9 +257,7 @@ const Helpers = {
 
     const brushDomain = defaults({}, targetProps.brushDomain, domain);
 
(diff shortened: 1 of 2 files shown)


becca-bailey:
Hi @Chris7, thanks for pushing up a fix! Apologies that it took me so long to review it. It looks like there are a couple of linter errors. Also, I know that this is connected to the VictoryZoomContainer bug, but are you able to provide me with a test case so I can try it out locally?

becca-bailey:
@Chris7 I'm going to close this for now due to inactivity, but we can re-open if necessary. Thanks!

Issue #2031, opened 2021-12-02
Event is not woking on android in react-native


Siva-Leo:
My Code 
```
<VictoryChart
        width={size.responsiveWidth(100)}
        domain={{ y: [200, 1000] }}
        domainPadding={{ y: 10 }}
        theme={VictoryTheme.material}
      >
        <LinearGradient x1="0%" y1="0%" x2="0%" y2="100%" id="myGradient">
          <Stop offset="0%" stopColor={colors.primaryLight} stopOpacity={0.3} />
          <Stop offset="60%" stopColor={colors.background} />
        </LinearGradient>

        <VictoryAxis
          crossAxis
          style={{
            grid: { strokeWidth: 0 },
            axis: { stroke: 0 },
            tickLabels: { opacity: size.opacityMedium },
            ticks: { stroke: 0 },
          }}
          tickFormat={(tick, index) => (index % 2 === 0 ? tick : null)}
        />
        <VictoryAxis
          dependentAxis
          style={{
            grid: { strokeWidth: 2, strokeDasharray: 0 },
            axis: { stroke: 0 },
            tickLabels: { opacity: size.opacityMedium },
            ticks: { stroke: 0 },
          }}
        />
        <VictoryArea
          data={data[2020]}
          x="month"
          y="customer"
          style={{
            data: {
              fill: 'url(#myGradient)',
              stroke: colors.primaryLight,
              strokeWidth: 3,
            },
          }}
          interpolation="cardinal"
        />
        <VictoryBar
          data={data[2021]}
          x="month"
          y="customer"
          dataComponent={
            <Bar
              events={{
                onClick: evt => alert(`(${evt.clientX}, ${evt.clientY})`),
              }}
            />
          }
          barWidth={() => 7}
          events={[
            {
              target: 'data',
              eventHandlers: {
                onClick: () => {
                  return [
                    {
                      target: 'data',
                      mutation: barProps => {
                        const fill = barProps.style && barProps.style.fill;
                        return fill === 'black' ? null : { style: { fill: 'black' } };
                      },
                    },
                  ];
                },
              },
            },
          ]}
          style={{ data: { fill: colors.primary } }}
        />
      </VictoryChart>
```
Expected behavior
chart bar bg-color toggle to black or alert popup
but got nothing

AashiqDurga:
Have you managed to get a workaround? 

I have the some issue with events not firing on android but works fine on iOs

grean:
Same here.
[...]

PR #2071, opened 2022-01-25
fix: android touch events

zibs:
This adjusts the markup of the native `VictoryContainer` so that it's always rendering its children within an `Svg` element. My hunch is that this seems to fix touch issues on Android because `react-native-svg`'s `Svg` is central to its own implementation of touch events. See [here](https://github.com/react-native-svg/react-native-svg/blob/develop/android/src/main/java/com/horcrux/svg/SvgView.java) to see methods like `hitTest` and `interceptsTouchEvent`. This is why previously proposed solutions like [this](https://github.com/FormidableLabs/victory/issues/2031#issuecomment-1011077399) and [this](https://github.com/FormidableLabs/victory/issues/2031#issuecomment-1020921557) seem to fix the issue.

Currently, the `View` below the main `Svg` is positioned absolutely on top of it, and then renders an empty `Svg` (portal), which I think is why touch events aren't captured on Android.

Tested by running the demo app included in this repo, and by adding various touch events (not included in PR) to various charts and see if they still work.

On iOS this causes no changes (as far as I can tell).

Concerns:
1. On web this would be essentially invalid markup, having an SVG contain html/jsx without declaring it as a `foreignobject` first. However, I don't seem to encounter any errors, warnings, or complaints in doing this here, but just want to flag that.
2. Portals! I'm actually not sure how they work fully on native yet and want to continue investigating there next.

### DIFF
--- a/packages/victory-native/src/components/victory-container.js
+++ b/packages/victory-native/src/components/victory-container.js
@@ -158,10 +158,10 @@ export default class extends VictoryContainer {
           {title ? <title id="title">{title}</title> : null}
           {desc ? <desc id="desc">{desc}</desc> : null}
           {children}
+          <View style={portalDivStyle} pointerEvents="box-none">
+            <Portal {...portalProps} ref={this.savePortalRef} />
+          </View>
         </Svg>
-        <View style={portalDivStyle} pointerEvents="box-none">
-          <Portal {...portalProps} ref={this.savePortalRef} />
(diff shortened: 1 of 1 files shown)


becca-bailey (approved):
This makes sense to me! I'm also a little confused about why the view works better inside the SVG container, but if it work this is good with me. 👍

PR #2077, opened 2022-01-29
feat: Add VictoryNativeBrushLine component

### LINKED ISSUE #2072: VictoryBrushLine in React Native
# Feature Requests

### Checklist

- [v] I've read through the [Docs](https://formidable.com/open-source/victory/docs) and [Guides](https://formidable.com/open-source/victory/guides) to make sure this functionality doesn't already exist

- [v] I've searched open issues to make sure I'm not opening a duplicate issue

### Description

I want to use "Paralle Bursh Axes" in my react native app.
However, victory-native does not include "VictoryBrushLine".
So, please let me know you guys have plan to add "VictoryBrushLine" in react native library.

And i've tried to use "VictoryBrushContainer" to make same function as the example. 
However, I cannot use multiple container in one chart. If there is any method to use multiple container, please let me know.
Thanks for the great and awesome chart!!!

zibs:
This brings `VictoryBrushLine` parity to `victory-native`.

The implementation is a combination of techniques used in `victory-native`’s `VictoryContainer`, `Portal`, and the standard way of bringing components to RN.

Ultimately we still wrap the component with `wrapCoreComponent`, but before doing so we:

1. Extend the the component and override its render method so that it renders a valid `react-native-svg` component (much like how `Portal` is done).
2. Remap the relevant `static defaultEvents` to their native equivalents e.g `onTouchStart -> onMouseDown` (I feel like I'm kind of hacking this, and there's a better way to do so?)
3. Create a set of `PanResponders` like in the native `VictoryContainer` that will respond to the touch start/move/end events that are attached to each brush line.

Adds typings/smoke-test
Adds docs/screen to RN demo app

Resolves #2072 

This actually requires #2071 to function correctly on Android due to the touch events issue.

https://user-images.githubusercontent.com/1131641/151640395-6a756a7c-5e52-497f-a9ed-a564011e139c.mov

https://user-images.githubusercontent.com/1131641/151640399-463b7ea9-cae2-44d5-a08f-f719673655d3.mov


becca-bailey (approved):
This looks good to me! I think the only comment I have is that eventually we want to move away from extending classes for `victory-native` components, because this makes it harder to change the base components. In my perfect world, we would use composition to pass in alternate props for `victory-native` rather than extending a class. HOWEVER, this is outside the scope for this change, and we can continue using this syntax until we're able to prioritize updating this pattern across the board.

PR #2082, opened 2022-02-04
Bugfix/2078 

dlcartright:
### Description:

Passed missing 'title' and 'desc' props to VictoryContainer component from VictoryChart component. Added functionality for users to pass other attributes to the container SVG that pass the whitelist test (defined in victory/packages/victory-core/src/victory-util/user-props.js).

Currently the whitelist contains 2 arrays called "startsWith" and "exactMatch".  Any props will be run against these two to see if they start with any of the items in the startsWith array or they match exactly with items from the exactMatch array.

### Testing:
Start this branch with `yarn start` and navigate to the [Chart example](http://localhost:3000/#/chart).

#### Test 1
1. Inspect the first chart in DevTools, you will see an attribute added to the SVG called "data-some-user-prop" as well as an "aria-label".  
2. Any of other attributes passed into VictoryChart (that are allowed in the whitelist) will be added here as well.  So, if you wish, add additional attributes to the example and inspect to see that they carried over.
UserAttributes

#### Test 2
1. Using the browser extension called _Accessibility Insights for Web_ run a **Fast Pass** on the chart example page.
2. The first 4 charts will be accessible (because they have a title attribute added to them). The others will be outlined in red and show the error below.  It is up to the user to add the title to their charts.

ChartsMissingTitle


packages/victory-core/src/victory-container/victory-container.js
```diff
@@ -160,7 +166,7 @@ export default class VictoryContainer extends React.Component {
           ref={this.saveContainerRef}
           {...this.getOUIAProps(props)}
         >
-          <svg {...svgProps} style={svgStyle}>
+          <svg {...svgProps} {...safeUserProps} style={svgStyle}>
```
becca-bailey, line 169: Not a blocker - would it make more sense just to add these safe user props to the `svgProps` object so we're not passing in more than one props object? Sometimes prop overrides ad defaults can get a bit confusing when we destructure multiple objects here.

becca-bailey:
Hi @dlcartright, I think I'm going to create a feature branch for this work, and we can merge this in-progress work to the feature branch. I think that will make it easier for us to divide up the work for the implementation in other components. Does that work for you?

dlcartright:
@becca-bailey , yes. That works. Sorry I haven't been able to circle back to this in a while. I'll take a look today.

packages/victory-core/src/victory-container/victory-container.js
dlcartright: You got it!

dlcartright:
Consolidated work in a new branch (allow-user-props-foundation). Will continue work from that branch.

PR #2089, opened 2022-02-09
Patch missing text component type

youPickItUp:
Hello!

I noticed the docs and typescript were missing the `textComponent` prop on `VictoryLabel` which I needed to overwrite.
Here is PR fixing that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 36.2.1 (2022-01-28)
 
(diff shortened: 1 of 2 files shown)


becca-bailey:
Thanks so much @youPickItUp ! I'll get this merged in, and it should go out with the next release.

becca-bailey:
It looks like the chromatic and doc deploy failures are happening because Github isn't able to run these actions from a PR opened from a fork. I'm going to merge this to a Victory branch first just to make sure these checks are running.

Issue #2094, opened 2022-02-14
Fix `compile-ts` so it catches typescript errors and add check to CI


becca-bailey:
In `package-scripts.js`, there is a `nps compile-ts` command. However, this isn't currently catching TS build errors, like the one that was created by #2058. We should make sure this compile script shows us real-time TS type errors and add a TS check step to CI. 

Steps to reproduce:
1. Uncomment a non-implemented type in `victory/src/index.d.ts`
2. run `nps compile-ts`
3. See that there are no errors

matt-hernandez:
@becca-bailey Did a little poking around. The core of the issue, from what I can tell, is that the typescript compiler doesn't include `.d.ts` files in the compilation process at all. So a `.d.ts` file could have any number of errors and the compiler would never complain. I was hoping that ESLint would cover this, but no luck so far.

Did some toying around and a general solution I found was to change `packages/victory/src/index.d.ts` to `packages/victory/src/index.ts` and adding these two lines to `tsconfig.json`

```json
{
  "compilerOptions": {
    ...
    "declaration": true,
    "emitDeclarationOnly": true
  },
  ...
}
```

With a general `.ts` file, `tsc` kicks in and flags the non-implemented imports. And with the two additional compiler options, it also makes sure to ignore duplicate identifiers from other files. Keeps the focus on whether the imports exist, and only on that.

Since this is a structural change that might require review, I'm posting this comment here. If you think this is okay, I already have the repo forked and can open an PR

becca-bailey:
Thanks so much @matt-hernandez ! You saved me a couple hours of digging today to figure all of this out myself. Feel free to open your PR so I can check it out!

becca-bailey:
Once we have the ts-compile script (or another check script) working as expected, I can handle any updates in CI.

matt-hernandez:
#2107 was made

It required a bit more housekeeping adjustments than I expected, but the overall crux of the idea is there. I'll be on standby if you need me

matt-hernandez:
@becca-bailey I may have something! I looked over #2084 which was the issue that first raised the bug of unimplemented types. That issue specifies that a typescript project which uses Victory will only fail to compile if `skipLibCheck` in the project's `tsconfig.json` is `false`.

So I did a little experiment and set `tsconfig.json` of the Victory repo to this

```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "outDir": "demo/dist",
    "rootDir": ".",
    "module": "es6",
[...]

PR #2107, opened 2022-02-15
Fix TS build 2094

matt-hernandez:
## Description:

In the main code of the project, all packages have `index.d.ts` files. `skipLibCheck` is set to `true` in the project `tsconfig.json`, which means that unimplemented types can slip through these files without causing a compile error. These errors can then show up after a build release, when users install Victory into their TypeScript projects and set `skipLibCheck` to `true`.

### How the fix is made

- In `tsconfig.json`
```diff
{
  "compilerOptions": {
    "baseUrl": ".",
    "outDir": "demo/dist",
    "rootDir": ".",
    "module": "es6",
    "target": "es5",
    "jsx": "react",
    "sourceMap": true,
    "moduleResolution": "node",
+   "typeRoots": [
+     "./node_modules/@types"
+   ],
+   "types": ["react", "lodash"],
    "allowSyntheticDefaultImports": true,
    "forceConsistentCasingInFileNames": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "noImplicitAny": true,
    "noUnusedLocals": true,
    "strictNullChecks": true,
    "strict": true,
    "suppressImplicitAnyIndexErrors": true,
    "paths": {
      "@packages/*": ["packages/*"]
    },
-   "skipLibCheck": true
+   "skipLibCheck": false
  },
  "include": [
    "**/src/*",
    "demo/ts/*"
  ],
  "exclude": [
      "node_modules",
      "**/*.spec.ts"
  ]
}
```
- Types for react and lodash (which is used in `demo`) are explicitly included

### Testing:

1. In `packages/victory/src/index.ts`, uncomment any unimplemented type Screen Shot 2022-02-14 at 6 32 30 PM
2. Run the command `nps compile-ts` Screen Shot 2022-03-01 at 2 00 51 PM

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -8,6 +8,10 @@
     "jsx": "react",
     "sourceMap": true,
     "moduleResolution": "node",
+    "typeRoots": [
+      "./node_modules/@types"
+    ],
+    "types": ["react", "lodash"],
     "allowSyntheticDefaultImports": true,
     "forceConsistentCasingInFileNames": true,
     "noImplicitReturns": true,
@@ -20,7 +24,7 @@
     "paths": {
       "@packages/*": ["packages/*"]
     },
-    "skipLibCheck": true
+    "skipLibCheck": false
   },
   "include": [
     "**/src/*",


matt-hernandez:
So this solution I came up involved a lot more changes around the configuration and housekeeping of this project than I anticipated. But in its current state, it will complete `nps lerna-dry-run` without error. If there are questions around the reasons for any of my changes, I'll be watching this PR

becca-bailey:
Hi @matt-hernandez, thanks for opening this PR! Apologies that I haven't gotten a chance to review it yet. I'm going to do a bit more investigation, and I might ask some other contributors to weigh in her as well.

matt-hernandez:
@becca-bailey No problem. I've been experimenting in my free time to see if there's a fix that involves a little less overhaul, but having a surprising amount of trouble. But I think the core of the idea is the same, a regular `.ts` somewhere in the package chain to kick off type checking that will break if an unimplemented type is exported.

If someone can find a way to do that with less changes, I'm all for it.

matt-hernandez:
See latest comment in #2094

matt-hernandez:
Branch updated with smaller fix

becca-bailey:
Approved! Thanks @matt-hernandez!

matt-hernandez:
If there was a party parrot reaction, I would certainly click it @becca-bailey. Thank you!

Issue #2040, opened 2021-12-20
Cannot add data-testid or id for tests


RobelTekle:
- Victory version 36.1.0
- React version 17.0.2

I'm trying to add some `data-testid` to Victory components in order to do some tests with Cypress. I have tried with several components: VictoryChart, VictoryBar, VictoryContainer. They never pass down the `data-testid` even if this prop is accepted by TypeScript. 

Could be very useful to have this feature, or at list to be able to passe down an `id`

becca-bailey:
Hi @RobelTekle ! Victory currently only passes expected props down to child components from the top-level components. If you want a way around this, you can use component props like this:

```jsx
<VictoryBar dataComponent={<Bar pathComponent={<path data-testid="bar" />} />} />
```

It's a bit annoying, and we should investigate passing arbitrary props down so you don't have to use this workaround.

pranay101:
Hi, I am new to open source and I would like to work on this issue. Please guide me further on what needs to be done.

becca-bailey:
Hi @pranay101, I believe this issue is being worked on by another contributor, but feel free to check out our good first issues board if you would like to find a different issue to work on. Feel free to tag me directly if there are specific things you need help with! https://github.com/FormidableLabs/victory/projects/2

PR #2151, opened 2022-03-05
Added functionality to be able to pass user props to components. Safe…

dlcartright:
### Description:

Passed missing 'title' and 'desc' props to VictoryContainer component from VictoryChart component. Added functionality for users to pass other attributes to the container SVG that pass the whitelist test (defined in victory/packages/victory-core/src/victory-util/user-props.js).  Safe-list currently allows props starting with 'aria-' and 'data-'.  VictoryChart, VictoryGroup, VictoryStack, and VictoryContainer have all been wired up to allow these user props. More components to follow.

Currently the whitelist contains 2 arrays called "startsWith" and "exactMatch".  Any props will be run against these two to see if they start with any of the items in the startsWith array or they match exactly with items from the exactMatch array.

### Testing:
Start this branch with `yarn start` and navigate to the [Chart example](http://localhost:3000/#/chart).

#### Test 1
1. Inspect the first chart in DevTools, you will see an attribute added to the SVG called "data-some-user-prop" as well as an "aria-label".  
2. Any of other attributes passed into VictoryChart (that are allowed in the whitelist) will be added here as well.  So, if you wish, add additional attributes to the example and inspect to see that they carried over.
UserAttributes

#### Test 2
1. Using the browser extension called _Accessibility Insights for Web_ run a **Fast Pass** on the chart example page.
2. The first 4 charts will be accessible (because they have a title attribute added to them). The others will be outlined in red and show the error below.  It is up to the user to add the title to their charts.

ChartsMissingTitle


dlcartright:
Will be making additional PRs into this branch for the remaining components to break up the work and keep the PRs manageable.

packages/victory-group/src/victory-group.js
```diff
@@ -88,17 +89,21 @@ const VictoryGroup = (initialProps) => {
     name
   ]);
 
+  const userProps = React.useMemo(() => UserProps.getSafeUserProps(initialProps), [initialProps]);
```
becca-bailey, line 92: I'm not sure whether this `useMemo` is doing any good here. Since `initialProps` isn't stored in a state or ref value, I believe `useProps` will be re-calculated on every render. There is a fairly good explanation here. https://github.com/FormidableLabs/victory/issues/2040

For now, I wouldn't stress too much about memoizing this value since it isn't a very expensive function call. Unless you can see a performance difference, let's just remove the `useMemo` for now.

packages/victory-core/src/victory-util/add-events.js
```diff
@@ -274,11 +275,14 @@ export default (WrappedComponent, options) => {
...
         ? this.getComponentProps(component, "parent", "parent")
         : {};
+
       if (parentProps.events) {
         this.globalEvents = Events.getGlobalEvents(parentProps.events);
         parentProps.events = Events.omitGlobalEvents(parentProps.events);
       }
-      return React.cloneElement(component, parentProps, children);
+      
+      const componentProps = { ...parentProps, ...parentProps.userProps };
```
becca-bailey, line 284: Where is `parentProps.userProps` coming from? Is this outdated, or are we still setting this somewhere in a `getComponentProps` function?

becca-bailey:
I have a couple questions and comments here, but let's merge this work into `feature/user-props`, and you (or anyone else who picks up the work) can address these comments in a follow-up PR to the same feature branch. I will also change your other PRs to merge into the feature branch, rather than into this one.

PR #2161, opened 2022-03-14
Upgrade storybook to v6

becca-bailey:


becca-bailey:
This was previously showing some small chromatic differences (it looked like there was a minor difference in the way stories were rendering in v6). Chromatic didn't run for this PR, but if this comes back we can either revert this work or accept the minor chromatic changes

Issue #2083, opened 2022-02-04
Horizontal charts do not pan at the correct rate 


taylorbf:
### The Problem

When a chart is set to `horizontal` mode and is zoomed and panned, the panning rate is not 1:1 with mouse movement. 

This appears to occur only if the `height` and `width` of the chart are different.   I wonder if the panning rate is calculated relative to the `height` of the chart when it should be calculated according to the `width` of the chart for left-right panning.   

Suggestions or workarounds welcome.

### Reproduction

https://codesandbox.io/s/autumn-firefly-br1sm?file=/src/App.js

becca-bailey:
Hi @taylorbf, thanks for the reproduction! I am seeing these inconsistencies with the way panning is happening in each chart, and I'm uncertain about whether this is intended behavior. I'm not sure whether this issue will be able to be addressed individually, but we are planning to rewrite the zooming behavior sometime in the near future, and I hope that this rewrite will fix some of these smaller inconsistencies. Here's the current ticket where we are tracking issues related to zooming: #2111

becca-bailey:
I'm going to close this ticket for now and roll this issue into the rewrite ticket.

TabithaMaudMM:
@becca-bailey I think I have a fix for this bug. Are you currently accepting PRs? 

It looks like getDomainScale is setting the range incorrectly. 
```
  getDomainScale(domain, scale, axis, horizontal) {
    const axisDomain = Array.isArray(domain) ? domain : domain[axis];
    const [from, to] = axisDomain;
    const otherAxis = axis === "x" ? "y" : "x";
    const range = horizontal ? scale[otherAxis].range() : scale[axis].range();
    const plottableWidth = Math.abs(range[0] - range[1]);
    return plottableWidth / (to - from);
  },
```

const range does not need to switch the axis - scale already handles that:
```
  const scale = {
    x: Scale.getBaseScale(props, "x")
      .domain(domain.x)
      .range(props.horizontal ? range.y : range.x),
    y: Scale.getBaseScale(props, "y")
      .domain(domain.y)
      .range(props.horizontal ? range.x : range.y)
  };
```

 So range would just need to be:
```
 const range =  scale[axis].range();
```

This fixes horizontal panning and I don't see any other issues. We need this for a project we're working on, I know this is part of a bigger rewrite ticket but if you're accepting PRs could this be a temporary fix until that bigger ticket is completed?

becca-bailey:
[...]

PR #2179, opened 2022-04-04
Fix Horizontal charts panning at the incorrect rate

TabithaMaudMM:
This is in regards to this bug reported for horizontal graphs: https://github.com/FormidableLabs/victory/issues/2083

### The Problem
When a chart is set to horizontal mode and is zoomed and panned, the panning rate is not 1:1 with mouse movement.

It looks like getDomainScale is setting the range incorrectly.
```
  getDomainScale(domain, scale, axis, horizontal) {
    const axisDomain = Array.isArray(domain) ? domain : domain[axis];
    const [from, to] = axisDomain;
    const otherAxis = axis === "x" ? "y" : "x";
    const range = horizontal ? scale[otherAxis].range() : scale[axis].range();
    const plottableWidth = Math.abs(range[0] - range[1]);
    return plottableWidth / (to - from);
  },
```

const range does not need to switch the axis - scale already handles that:
```
  const scale = {
    x: Scale.getBaseScale(props, "x")
      .domain(domain.x)
      .range(props.horizontal ? range.y : range.x),
    y: Scale.getBaseScale(props, "y")
      .domain(domain.y)
      .range(props.horizontal ? range.x : range.y)
  };
```

 So range would just need to be:
```
 const range =  scale[axis].range();
```

This appears to fix horizontal panning! Everything else appears to still be working properly as well.

### DIFF
--- a/packages/victory-zoom-container/src/zoom-helpers.js
+++ b/packages/victory-zoom-container/src/zoom-helpers.js
@@ -138,7 +138,7 @@ const RawZoomHelpers = {
       ? newDomain.map((val) => new Date(val))
       : newDomain;
   },
-  // eslint-disable-next-line max-params
+
   getDomainScale(domain, scale, axis) {
     const axisDomain = Array.isArray(domain) ? domain : domain[axis];
     const [from, to] = axisDomain;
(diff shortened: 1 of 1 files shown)


becca-bailey:
Thanks for the fix @TabithaMaudMM! I'll run all the CI checks, then we should be able to get this change merged in.

becca-bailey:
@TabithaMaudMM It looks like the `format` script might be returning an error in CI. Are you able to run `nps format` on your branch?

TabithaMaudMM:
Thanks @becca-bailey! Just ran the yarn nps format and pushed up the changes 👍

PR #2192, opened 2022-04-14
Allow additional props from user, such as `aria-label` and `data-testid`

### LINKED ISSUE #2040: Cannot add data-testid or id for tests

becca-bailey:
- Allows `title` and `desc` to be passed to Victory components
- Allows `data-testid` and `aria-*` props to be passed through using an allow list

@dlcartright I closed your original PRs, and opened a new one with just the component changes and tests.

Closes #2040


becca-bailey:
I'm going to merge this for now - there is still some outstanding work for primitive components and nested components like `Line` and `Bar`.

Issue #2195, opened 2022-04-14
Replace Enzyme with React Testing Library


becca-bailey:
In addition to some of the general issues with our tests being a little slow and out of date, we will need to replace Enzyme as our method of rendering components in tests if we want to upgrade to React 18+. 

We should re-write our existing tests using Jest + React Testing Library.

Related issue: #2187 

See also: https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl

## Completed packages

- [x] victory
- [x] victory-area
- [x] victory-bar
- [x] victory-box-plot
- [x] victory-brush-container
- [x] victory-candlestick
- [x] victory-chart
- [x] victory-core
- [x] victory-errorbars
- [x] victory-group
- [x] victory-histogram
- [x] victory-legend
- [x] victory-line
- [x] victory-pie
- [x] victory-scatter
- [x] victory-selection-container
- [ ] victory-shared-events (this might be going away soon, so it's a lower priority test suite to convert)
- [x] victory-stack
- [x] victory-tooltip
- [x] victory-voronoi

## General tips and pointers

- We can use `querySelector` in a test to test attributes on SVG elements. 
- We can also use `getByRole` or `getByTestId` to isolate specific nodes. #2191 should allow us to add `data-testid` to our test components, but there may be some primitive components that need to be set up to work with these user-provided props.
- We can replace some of the test assertions with `.toMatchInlineSnapshot`, provided that the snapshot is focused (< 10 lines) and easy to read. 
- If we want to test the data that is being passed to the SVG component, there are some workarounds we can use, such as giving our test component a property that outputs the rendered data as JSON. Here is an example:

```jsx
it("renders data ordered by value of sortKey, if given", () => {
      const data = [{ t: 0 /*x: 10, y: 1*/ }, { t: 1 /*x:  9, y: 1*/ }];
      const { container } = render(
        <VictoryLine
          data={data}
          sortKey="t"
          x={({ t }) => 10 - t}
          y={() => 1}
          dataComponent={
            <Curve data-json={(props) => JSON.stringify(props.data)} />
          }
        />
      );

      const line = container.querySelector("path");
      const renderedData = JSON.parse(line.getAttribute("data-json")).map(
        ({ t }) => t
      );

      expect(renderedData).toEqual([0, 1]);
    });
```
Long-term, we might want to figure out how to test data transformations in a different place.

heythisispaul:
@becca-bailey I can start work on victory-bar.

[...]

PR #2196, opened 2022-04-14
Add jest and react testing library tests for victory-line and victory-area

becca-bailey:
This is some initial work to switch over tests from Karma + enzyme to Jest + RTL. This adds new tests, but doesn't delete the original tests just yet.

Completed packages:

- [x] victory-line
- [x] victory-area

Related issue: #2195 

To run jest tests:
```sh
yarn nps jest
```
OR 
```sh
yarn nps jest.native
```

To pick up code changes in Victory packages, this test command will need to be run concurrently with `yarn nps watch`.


package.json
```diff
@@ -31,7 +31,8 @@
     "storybook": "nps storybook.server",
     "build-storybook": "build-storybook",
     "chromatic": "chromatic",
-    "chromatic-ci": "build-storybook && chromatic --ci --exit-zero-on-changes"
+    "chromatic-ci": "build-storybook && chromatic --ci --exit-zero-on-changes",
+    "test": "jest --config=jest-config.js"
```
becca-bailey, line 35: @ryan-roemer Right now, we are able to run jest tests using the `jest` command. However, we need to manually rebuild modules using `nps build-package-libs` every time there are code changes in a package. Re-building before running each test is what has been causing the current test suite to be so slow. You added the `TEST_MODULE` flag last year that will build one package in isolation, but this doesn't work for automatically re-building changed files if we are running jest in watch mode. Do you know of any tooling that could handle this better?

ryan-roemer: So, there are two things really:

1. Build **all** lib and test code when running tests.
2. Build only test code when running tests. For lib code that means running a separate `watch` to rebuild or one-off build.

I think No. 2 is the way to go.

And for Jest `TEST_MODULE` doesn't really apply. And for Karma, it has very little actual speed impact. The thing that _does_ now massively speed up Karma in my d3 branch is that I only use already build lib code and remove all transpiling and inclusion of source code.

ryan-roemer:
@becca-bailey Cool! And if the jest tests run fast enough on their own, then we maybe don't even need to heavily optimize for jest `watch`. (And jest watch should be fine watching with `<pkg>/es` same as `<pkg>/src` for auto-rerunning changes).

becca-bailey:
@ryan-roemer That's a good point - for now running `yarn nps watch` concurrently with `jest --watch` works well!

becca-bailey:
Any objections to merging in new tests as we go rather than waiting until they are all completed?

jest-config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  testEnvironment: "jsdom",
```
ryan-roemer, line 2: Do we want a single config or split configs for Jest + Native vs Jest + web (jsdom)?

test/jest/victory-native/native-helpers.test.js
```diff
@@ -1,4 +1,4 @@
-import NativeHelpers from "../../packages/victory-native/src/helpers/native-helpers";
+import NativeHelpers from "../../../packages/victory-native/src/helpers/native-helpers";
```
ryan-roemer, line 1: Note that this should already work without any prefix (and make integration with my branch easier) as:

```suggestion
import NativeHelpers from "victory-native/src/helpers/native-helpers";
```

jest-config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  testEnvironment: "jsdom",
+  moduleFileExtensions: ["ts", "tsx", "js", "jsx", "json", "node"],
+  testMatch: ["**/jest/**/?(*.)+(spec|test).[jt]s?(x)"],
+  testPathIgnorePatterns: ["victory-native"],
+  transform: {
+    "^.+\\.(js|jsx)$": [
+      "babel-jest",
+      { configFile: "./babel-native-config.js" }
```
becca-bailey, line 9: @ryan-roemer I split this out into separate configs again so we can run both `nps jest` and `nps jest.native` to run either the `victory-native` tests or the rest of the victory tests.

However, I'm running into some issues with the babel config. It works with the babel native config, but if I change this line to `babel-config.js` I get some cryptic errors. This file just has the same `@babel/react` and `@babel/env` presets that the `.babelrc` files have, but there's some piece of config in the metro config that must be required for `babel-jest` and I haven't figured out yet what it is. Any ideas about what could be wrong here?

jest-native-config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   moduleFileExtensions: ["ts", "tsx", "js", "jsx", "json", "node"],
-  testMatch: ["**/jest/**/?(*.)+(spec|test).[jt]s?(x)"],
+  testMatch: ["**/jest/victory-native/**/?(*.)+(spec|test).[jt]s?(x)"],
```
becca-bailey, line 3: We might want to move these tests to a completely separate folder, but for now I just configured the jest native tests to run tests only in this folder, and added an exclude path option to `jest-config.js` to exclude this folder.

babel-config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ["@babel/react", "@babel/env"]
+};
```
becca-bailey, line 3: This is the same config that is in all the package-specific `babelrc` files, but it's not working for test compilation.

jest-config.js
jpdriver: @becca-bailey i don't have full context here so this may not be the right answer at all; but it sounds like we may want the RN Babel preset here?

https://github.com/facebook/react-native/blob/main/template/babel.config.js
https://github.com/facebook/metro/tree/main/packages/metro-react-native-babel-preset

packages/victory-core/src/victory-util/user-props.js
```diff
@@ -48,6 +48,19 @@ const testIfSafeProp = (key) => {
...
+ * @param {any} value: maybe function value
+ * @param {Object} props: props object
+ * @returns {any} newValue
+ */
+const getValue = (value, props) => {
+  if (typeof value === "function") {
+    return value(props);
+  }
+  return value;
+};
```
becca-bailey, line 62: This is a minor change to the user-provided props that mimics the API of other Victory props and allows us to pass in a function that evaluates the props. 

For example:
```jsx
<VictoryLine aria-label={props => props.label} />
```

I added this small change to this PR because it was useful for testing, but I will make a note to add some additional documentation around this in a separate PR.

test/jest/victory-line/victory-line.test.js
```diff
@@ -0,0 +1,304 @@
...
+          }
+        />
+      );
+
+      const line = container.querySelector("path");
+      const renderedData = JSON.parse(line.getAttribute("data-json")).map(
+        ({ t }) => t
+      );
+
+      expect(renderedData).toEqual([0, 1]);
```
becca-bailey, line 181: This is kind of a hacky solution that allows us to test the rendered data, since we're no longer able to directly access props with Enzyme (for a good reason!). 

I might do a bit of research to see if there are better approaches to testing rendered data in data visualization, since an automated test suite can't easily test things like _is the line going up or down_ the same way we could test it visually.

becca-bailey: It's also worth asking whether we should still be testing things like data ordering at the integration test level, or whether we should just unit test the helper functions.

becca-bailey:
I have some outstanding questions on how we want to approach some of these tests, but since this PR is making minimal code changes, I am okay merging this as-is as a jumping off point for future work. From there, we can re-evaluate which tests are most useful and make changes as needed.

PR #2204, opened 2022-04-26
Add victory-vendor package for d3 dependencies

becca-bailey:
This PR addresses the ESM-only package issues from d3 packages discussed in https://github.com/FormidableLabs/victory/issues/2124

We take a two-path approach in that ESM uses the **real** ESM-only packages and CJS uses our vendored (transpiled) package versions.

## Tasks

Stuff we still need to do:

- [x] Validate `victory-native` still works as expected.
- [x] Confirm storybook works as expected.
- [x] Docs: the build fails for `victory-vendor`
- [x] Docs: A build failure doesn't fail CI
- [x] Review and remove all TODOs.
- [x] Additional CHANGELOG.md notes.
- [x] Fix `it.skip()`-ed Karma tests.
- [x] Our `yarn nps lint.docs` command does not presently run in CI. It has a unique constraint in that both the root and `docs` dependencies need to be installed for it to succeed. We can either fix / enable here or maybe ticket for later.

## Work

New stuff:

- Add new `victory-vendor` package.
- Added a super-simple Node.js test to flush out future ESM issues faster. See #1965

Chores:

- Upgrade various babel deps that needed to be modernized.
- Refactor and simplify babel and webpack configs.
- Added more helper development commands in CONTRIBUTING.md

Fixes / infra:

- Shore up our import strategy for Jest and Karma. Now we _only_ use built lib code in tests (ESM in Karma, CJS in Jest). Unfortunately this touches a *lot* of files and we end up with a big PR.

## Release notes

We'll be publishing a new package `victory-vendor`, so I'm not sure if there are extra kinks in the release script there.

Also, for this release, since we've got a new vendor dependency scheme in the package, there are potentially some unforeseen issues that we won't be able to fully suss out until we fully release.

PR #2216, opened 2022-05-04
jest and rtl unit tests for bar components

heythisispaul:
This is part of the effort being recorded in #2195. This introduces jest/RTL unit tests for the `victory-bar` components.

This is only additive, copies over some of the existing tests and utils and updates them as needed/makes sense.

Give them a whirl with:

```
yarn nps jest
```


heythisispaul:
Still need to do the `geometry-helper-methods` tests that are with in the `victory-bar` client testing folder, but wanted to open this up to start getting some eyes/feedback.

test/jest/victory-bar/bar.spec.js
```diff
@@ -0,0 +1,83 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import * as d3Scale from "d3-scale";
```
ryan-roemer, line 3: As an FYI, PR #2204 is going to bring in our new vendoring approach to d3 libraries. If that one lands first, you'll need to update this (and similar imports except `d3-voronoi`) to:

```suggestion
import * as d3Scale from "victory-vendor/d3-scale";
```

If this PR lands first, then no changes are needed!

test/jest/victory-bar/victory-bar.spec.js
```diff
@@ -0,0 +1,257 @@
...
+
+describe("components/victory-bar", () => {
+  describe("default component rendering", () => {
+    it("accepts user props", () => {
+      const { container } = render(
+        <VictoryBar data-testid="victory-bar" aria-label="Chart" />
+      );
+
+      const svgNode = container.querySelector("svg");
+      expect(svgNode.getAttribute("data-testid")).toEqual("victory-bar");
```
becca-bailey, line 17: Now that we're using react testing library, we can change this to something like `expect(screen.getByTestId('victory-bar')).toBeDefined()`;

test/jest/victory-bar/victory-bar.spec.js
```diff
@@ -0,0 +1,257 @@
...
+      expect(svgNode.getAttribute("aria-label")).toEqual("Chart");
```
becca-bailey, line 18: Same here, we can use `screen.getByLabel('Chart')`

test/jest/victory-bar/victory-bar.spec.js
```diff
@@ -0,0 +1,257 @@
...
+      const bars = container.querySelectorAll("path");
+      expect(bars.length).toEqual(4);
+    });
+
+    it("renders each bar as a rectangle", () => {
+      const { container } = render(<VictoryBar />);
+      const barCommandStrings = Array.from(
+        container.querySelectorAll("path")
+      ).map((bar) => bar.getAttribute("d"));
+      barCommandStrings.forEach(svgExpectations.expectIsRectangular);
```
becca-bailey, line 45: This is a personal preference, but as we're making these changes I would prefer not to continue to nest test expectations themselves inside helpers. Something like `expect(isRectangular(bar)).toBeTrue()` is a little more readable and re-usable in multiple test files IMO.

test/jest/victory-bar/victory-bar.spec.js
```diff
@@ -0,0 +1,257 @@
...
+
+    it("renders bars for array-shaped data", () => {
+      const data = range(20).map((i) => [i, i]);
+      const { container } = render(<VictoryBar data={data} x={0} y={1} />);
+      const bars = container.querySelectorAll("path");
+      expect(bars).toHaveLength(20);
+    });
+
+    it("renders bars for deeply-nested data", () => {
+      const data = range(40).map((i) => ({ a: { b: [{ x: i, y: i }] } }));
```
becca-bailey, line 100: Tiny nitpick - we can probably render fewer bars here. I don't know whether that will have any impact on speed, but we're not getting any benefit to rendering 40 bars here, the test would work just as well if there were 4.

test/jest/victory-bar/victory-bar.spec.js
```diff
@@ -0,0 +1,257 @@
...
+      const { container } = render(
+        <VictoryBar data={data} x="a.b[0].x" y="a.b[0].y" />
+      );
+      const bars = container.querySelectorAll("path");
+      expect(bars).toHaveLength(40);
+    });
+
+    it("renders bars values with null accessor", () => {
+      const data = range(30);
```
becca-bailey, line 109: Same as above

test/jest/victory-bar/victory-bar.spec.js
```diff
@@ -0,0 +1,257 @@
...
+    });
+  });
+
+  describe("accessibility", () => {
+    it("adds an aria role to each bar in the series", () => {
+      const data = range(20).map((y, x) => ({ x, y }));
+      const { container } = render(<VictoryBar data={data} />);
+
+      Array.from(container.querySelectorAll("path")).forEach((bar) => {
+        expect(bar.getAttribute("role")).toEqual("presentation");
```
becca-bailey, line 233: Rather than iterating through each bar, you could do something like `expect(screen.queryByRole('presentation')).toHaveLength(20)`

test/svg-test-helper.js
```diff
@@ -80,3 +82,88 @@ export const calculateD3Path = (props, pathType, index = 0) => {
...
+};
+
+const exhibitsShapeSequence = (commandString, shapeSeqeuence) => {
+  const commands = parseSvgPathCommands(commandString);
+  return commands.every(
+    (command, index) => command.name === shapeSeqeuence[index]
+  );
+};
+
+export const svgExpectations = {
```
becca-bailey, line 120: We should update this syntax a bit and export one function at a time rather than exporting an object with multiple functions. I'm not sure if something like tree shaking matters in test utils, but we have been updating the rest of our utils to follow this convention. 

If we want them to be scoped, we can just create a different file that exports each function individually, and then we could do something like `import * as svgHelpers from '../svg-helpers'`.

test/svg-test-helper.js
```diff
@@ -80,3 +82,88 @@ export const calculateD3Path = (props, pathType, index = 0) => {
...
+  /**
+   * Asserts that the provided command string produces a rectangular shape.
+   *
+   * @param {string} commandString - the "d" attribute of a `path` element.
+   * @returns {void}
+   */
+  expectIsRectangular(commandString) {
+    const isBar = exhibitsShapeSequence(commandString, RECTANGULAR_SEQUENCE);
```
becca-bailey, line 128: As noted above, let's just export `isBar` here instead of including the test expecation.

becca-bailey:
@heythisispaul Thanks so much for doing this work! I left a few comments, but they're all pretty small fixes. My general feedback is that we can probably do a little more to update the syntax by using some of the specific RTL testing functions to cut down on using `querySelector` or `.getAttribute(...)`. We can also modernize some of the helpers a bit as well!

heythisispaul:
@becca-bailey Thanks for taking a look! I'll start working on those changes, I agree with a lot of your feedback. Most of this just came from me sort of copying the existing tests and trying to preserve them, rather than modernizing them.

I'll apply a little more creative license moving forward 😅

test/svg-test-helper.js
```diff
@@ -110,60 +110,54 @@ export const getPathCommandsFromContainer = (container) => {
...
 
-const exhibitsShapeSequence = (commandString, shapeSeqeuence) => {
+export const exhibitsShapeSequence = (commandString, shapeSeqeuence) => {
   const commands = parseSvgPathCommands(commandString);
   return commands.every(
     (command, index) => command.name === shapeSeqeuence[index]
   );
 };
 
-export const svgExpectations = {
```
heythisispaul: Just getting rid of this entirely so utilize can just be used in assertions within test files.

test/svg-test-helper.js
```diff
@@ -110,60 +110,54 @@ export const getPathCommandsFromContainer = (container) => {
...
+/**
+ * Retrieve the raw svg height of a bar.
+ *
+ * @param {string} commandString - The "d" attribute of a `path` element.
+ * @returns {Number}             - The height of the bar in svg units.
+ */
+export const getBarHeight = (commandString) => {
+  const commands = parseSvgPathCommands(commandString);
 
-export const getShapeUtils = {
```
heythisispaul: Functions are just being exported from this file directly, no longer grouped under objects.

becca-bailey:
One more small thing - I had named the other new tests with `.test.js` instead of `.spec.js` because that seems to be the jest convention. I don't have strong opinions, but should we re-name these test files for consistency?

heythisispaul:
Yeah, I don't either but I do agree that they should all match. I'll update these ones, sorry I didn't notice they were different before.

test/jest/victory-bar/geometry-heper.methods.test.js
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    it("the left-most point is the 0th element, the right-most is the 1st", () => {
+      const c0 = circle(point(2, 3), 3);
+      const c1 = circle(point(1, -1), 4);
+      const [{ x: x0 }, { x: x1 }] = c0.intersection(c1);
+      expect(x0 <= x1).toBeTruthy();
+    });
+  });
+});
```
becca-bailey, line 130: There is a typo in the file name - geometry-helper-methods.test.js

becca-bailey (approved):
I found one typo, but go ahead and merge when you are ready!

PR #2218, opened 2022-05-05
fix typos in scale prop documentation

heythisispaul:
After brushing up a little more on how the `scale` prop works, I noticed a few tiny grammatical issues in the [scale prop documentation](https://formidable.com/open-source/victory/docs/common-props/#scale).


becca-bailey:
Good catch, thanks!

PR #2225, opened 2022-05-09
add box plot jest/rtl tests

heythisispaul:
Part of effort #2195, updating box-plot unit tests to use Jest and React Test Library. 

These tests can be run with the command

```
yarn nps jest
```


becca-bailey (approved):
Looks good to me!

PR #2228, opened 2022-05-10
Dev: Switch to ye olde yarn1

ryan-roemer:
- Get rid of yarn2 files
- Add `engines` restriction
- Add CONTRBUTING notes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,9 @@
     "url": "https://github.com/formidablelabs/victory.git"
   },
   "private": true,
+  "engines": {
+    "yarn": "^1.22.0"
+  },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -13,6 +13,9 @@
     "url": "https://github.com/formidablelabs/victory.git"
   },
   "private": true,
+  "engines": {
```
ryan-roemer, line 16: Note: Since this is a private package.json that isn't published, should be safe for us. (We wouldn't want to restrict yarn engines on like our published packages...)

becca-bailey (approved):

PR #2246, opened 2022-05-19
Convert victory and victory-chart tests to jest

becca-bailey:
+ Tests for victory and victory-chart
+ Adds user props to VictoryAxis


test/client/spec/victory-axis/victory-axis.spec.js
```diff
@@ -15,6 +15,15 @@ import { _approximateTextSizeInternal } from "victory-core/es/victory-util/texts
 
 describe("components/victory-axis", () => {
   describe("default component rendering", () => {
+    it("accepts user props", () => {
```
heythisispaul, line 18: Just curious - is there a reason you added this test in the `client/spec` test suite as well? I was under the impression these were going to get phased out once the jest ones were complete.

test/jest/victory-chart/helper-methods.test.js
```diff
@@ -0,0 +1,48 @@
...
+  const getVictoryLine = (props) => React.createElement(VictoryLine, props);
+  const getVictoryAxis = (props) => React.createElement(VictoryAxis, props);
+
+  describe("getChildComponents", () => {
+    const defaultAxes = {
+      independent: getVictoryAxis({}),
+      dependent: getVictoryAxis({ dependentAxis: true })
+    };
+    beforeEach(() => {
+      jest.spyOn(Log, "warn");
```
heythisispaul, line 17: Is this just to silence the logs during tests?

test/jest/victory-chart/victory-chart.test.js
```diff
@@ -0,0 +1,101 @@
...
+            offsetY={50}
+            data-cross-axis={(props) => props.crossAxis}
+            data-offset-x={(props) => props.offsetX}
+            data-offset-y={(props) => props.offsetY}
+          />
+        </VictoryChart>
+      );
+      const axis = screen.getByTestId("axis");
+
+      expect(axis.getAttribute("data-cross-axis")).toEqual("false");
```
scottrippey, line 68: We might want to consider using [jest-dom](https://github.com/testing-library/jest-dom) in the future so this can be:
```suggestion
      expect(axis).toHaveAttribute("data-cross-axis", "false");
```
Not necessarily because of the code difference, but more because the error messages if it fails.

scottrippey (approved):
Looks great!  Seems nice and easy to convert these tests, and I'm looking forward to getting my hands dirty with this soon.

test/client/spec/victory-axis/victory-axis.spec.js
becca-bailey: I added it here as a reminder to include this test when we convert the victory axis tests. If it had been a complex test case I couldn't have bothered, but I just needed some basic confirmation that this was working, and we haven't converted this component yet.

test/jest/victory-chart/helper-methods.test.js
becca-bailey: I think so? I had just copied it over from the previous test, but I'll try removing it and see what happens.

test/jest/victory-chart/victory-chart.test.js
becca-bailey: Yep, that syntax is slightly better.

test/jest/victory-chart/helper-methods.test.js
becca-bailey: Removing it doesn't make a difference, so I'll just get rid of this.

PR #2247, opened 2022-05-20
Feature/jest candlestick tests

heythisispaul:
Part of the effort behind #2195, this migrates the existing `victory-candlestick` unit tests to RTL and Jest.

### DIFF
--- a/packages/victory-candlestick/src/index.js
+++ b/packages/victory-candlestick/src/index.js
@@ -1,2 +1,3 @@
 export { default as VictoryCandlestick } from "./victory-candlestick";
 export { default as Candle } from "./candle";
+export * as helperMethods from "./helper-methods";
--- a/test/jest/victory-candlestick/helper-methods.test.js
+++ b/test/jest/victory-candlestick/helper-methods.test.js
@@ -1,0 +1,77 @@
+/* eslint max-nested-callbacks: 0 */
+import { helperMethods } from "victory-candlestick";
+import { range } from "lodash";
+import { fromJS } from "immutable";
+
+const immutableGetDataTest = {
+  createData: (x) => fromJS(x),
+  testLabel: "with immutable data"
+};
+const getDataTest = {
+  createData: (x) => x,
+  testLabel: "with js data"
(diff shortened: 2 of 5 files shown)


packages/victory-candlestick/src/index.js
```diff
@@ -1,2 +1,3 @@
 export { default as VictoryCandlestick } from "./victory-candlestick";
 export { default as Candle } from "./candle";
+export * as helperMethods from "./helper-methods";
```
becca-bailey, line 3: I don't think we need to include this export here - we should be able to change the import in the test from `victory-candlestick/es/helper-methods` to `victory-candlestick/lib/helper-methods`.

test/jest/victory-candlestick/helper-methods.test.js
```diff
@@ -0,0 +1,77 @@
...
+  testLabel: "with immutable data"
+};
+const getDataTest = {
+  createData: (x) => x,
+  testLabel: "with js data"
+};
+
+const { getData, getDomain } = helperMethods;
+
+[getDataTest, immutableGetDataTest].forEach(({ createData, testLabel }) => {
```
becca-bailey, line 17: I know this is how the original test file was set up, but since there are only a few tests here personally I would just copy/paste the test assertions and add an extra "describe" block for immutable data. What do you think @heythisispaul ?

heythisispaul: Yeah I'm good with that, I think that's a little more declarative of an approach.

heythisispaul: I ended up making this change, found it to be a lot of code duplication, made some more adjustments and pretty much just wound back up with something very similar to this implementation so I decided to just keep it. I'm definitely still open to shaking it up if you have any specific ideas though!

becca-bailey: No problem, we can keep it the same for now.

becca-bailey (approved):

PR #2249, opened 2022-05-20
Add victory core tests

becca-bailey:
This is part of the work for #2195.

* Copies existing spec files, runs codemod to change expectations
* Replaces enzyme with RTL

Note: This does not include the tests for `add-events`. These tests were the most complex to convert (there is a lot of mocking going on in the current test implementation), and this `addEvents` HOC is also one of my top priorities for refactoring out of the codebase. I personally am okay with letting these tests go for now, and then adding additional test coverage once we have a better solution for all this shared logic.

### DIFF
--- a/test/jest/rendering-utils.js
+++ b/test/jest/rendering-utils.js
@@ -1,0 +1,6 @@
+import { render } from "@testing-library/react";
+import * as React from "react";
+
+export const renderInSvg = (component) => {
+  return render(<svg>{component}</svg>);
+};
--- a/test/jest/victory-core/victory-label/victory-label.test.js
+++ b/test/jest/victory-core/victory-label/victory-label.test.js
@@ -1,0 +1,246 @@
+import React from "react";
+import { VictoryLabel, Log } from "victory-core";
+import { screen, fireEvent } from "@testing-library/react";
+import { renderInSvg } from "../../rendering-utils";
+
+describe("components/victory-label", () => {
+  it("accepts user props", () => {
+    renderInSvg(
+      <VictoryLabel
+        data-testid="victory-label"
+        aria-label="test-aria-label"
+        text="label"
+      />
+    );
+
+    expect(screen.getByTestId("victory-label")).toBeDefined();
+    expect(screen.getByLabelText("test-aria-label")).toBeDefined();
+  });
+
+  it("has expected content with render", () => {
+    const { container } = renderInSvg(<VictoryLabel text="such text, wow" />);
+    expect(container.querySelector("tspan").innerHTML).toMatchInlineSnapshot(
+      `"such text, wow"`
+    );
+  });
+
+  it("sets dx and dy for text element", () => {
+    const { container } = renderInSvg(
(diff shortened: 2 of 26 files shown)


test/jest/rendering-utils.js
```diff
@@ -0,0 +1,6 @@
+import { render } from "@testing-library/react";
+import * as React from "react";
+
+export const renderInSvg = (component) => {
```
becca-bailey, line 4: Now that we're not using enzyme's shallow renderer, we needed a way to prevent warnings about unrecognized SVG elements in the DOM. Rendering SVG elements in an SVG container fixes this problem.

test/jest/victory-core/victory-label/victory-label.test.js
```diff
@@ -0,0 +1,246 @@
...
+  });
+
+  it("sets dx and dy for text element", () => {
+    const { container } = renderInSvg(
+      <VictoryLabel dx={30} dy={30} text="such text, wow" />
+    );
+    const output = container.querySelector("text");
+    expect(output.getAttribute("dx")).toEqual("30");
+    // dy = props.dy + (capHeight(0.71) / 2 + (0.5 - length(1) / 2) * lineHeight(1)) * fontSize(14);
+    expect(output.getAttribute("dy")).toBeNull();
```
becca-bailey, line 34: I'm not sure why this is null when dy is one of the attributes being tested? But I'm guessing it was an intentional change at some point.

test/jest/victory-core/victory-util/data.test.js
```diff
@@ -0,0 +1,382 @@
...
+            three: 3,
+            red: 4,
+            cat: 5
+          });
+        });
+      });
+    });
+  });
+
+  [dataTest, immutableDataTest].forEach(({ createData, testLabel }) => {
```
becca-bailey, line 62: This is the same pattern I commented on @heythisispaul 's PR - in the future I would like to avoid creating `describe` blocks inside a `forEach`, but for now we'll keep it while we copy over these tests.

test/jest/victory-core/victory-primitives/curve.test.js
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { Curve } from "victory-line";
+import { renderInSvg } from "../../rendering-utils";
+import * as d3Scale from "victory-vendor/lib/d3-scale";
```
ryan-roemer, line 4: This should be `victory-vendor/d3-scale` unless we're having a Jest problem with it using the ESM version like @scottrippey was diagnosing today? (Technically in Node.js you shouldn't be able to specify `victory-vendor/lib/*`, but Jest doesn't use real Node.js require's)

test/jest/rendering-utils.js
scottrippey: THANK YOU!  I was stubbing `console.error` for a while just to silence these errors; glad you've got a solution!

scottrippey (approved):

PR #2252, opened 2022-05-23
Test migration: victory-tooltip, victory-voronoi

scottrippey:
Copied tests and migrated to RTL + Jest.

Most tests translated verbatim, a few `prop` assertions had to be dropped.


test/jest/victory-tooltip/victory-tooltip.test.js
```diff
@@ -0,0 +1,53 @@
...
+    x: 0,
+    y: 0,
+    datum: { some: "object" },
+    index: 3,
+    active: true,
+    text: "such text, wow"
+  };
+  beforeAll(() => {
+    jest.spyOn(console, "warn").mockImplementation(() => {});
+    jest.spyOn(console, "error").mockImplementation(() => {});
```
becca-bailey, line 20: I just merged in my PR, so you should be able to update this to use the `renderInSvg` function if you're still seeing these unrecognized DOM element errors.

test/jest/victory-tooltip/victory-tooltip.test.js
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it("renders nothing when not active", () => {
+    render(<VictoryTooltip {...baseProps} active={false} />);
+    const output = screen.findByText(baseProps.text);
+    expect(output).toBeDefined();
```
becca-bailey, line 26: This might be how the original was, but should we add a test assertion here that confirms the tooltip isn't visible? This test assertion is exactly the same as the one on line 32.

test/jest/victory-tooltip/victory-tooltip.test.js
```diff
@@ -0,0 +1,53 @@
...
+  it("has expected text", () => {
+    render(<VictoryTooltip {...baseProps} />);
+    const output = screen.findByText(baseProps.text);
+    expect(output).toBeDefined();
+  });
+
+  it("renders a flyout and a label", () => {
+    const { container } = render(<VictoryTooltip {...baseProps} />);
+    const label = container.querySelector("text");
+    const flyout = container.querySelector("path");
```
becca-bailey, line 38: Nitpick: We can use the `flyoutComponent` and `labelComponent` props here to pass in components with a test id and use `findByTestId`. That way we can be more specific and confirm that these component we are finding are actually being rendered as the flyout and label, and not some other SVG child.

becca-bailey: If you go for that option, we will need to add the `getSafeUserProps` function to `Flyout` like I did for `VictoryLabel` in my last PR.

test/jest/victory-voronoi/victory-voronoi.test.js
```diff
@@ -0,0 +1,201 @@
...
+        ]
+      };
+      const { container } = render(<VictoryVoronoi {...props} />);
+
+      expect(container.querySelector("path").getAttribute("d")).toEqual(
+        calculateD3Path(props, "voronoi", 0)
+      );
+    });
+
+    it.skip("sorts data by sortKey prop", () => {
```
becca-bailey, line 58: This test is a weird one to convert. Because most Victory props can be accessor functions that can get prop values, I have a really hacky solution here that accesses the data prop and stringifies it in the DOM for RTL to access. https://github.com/FormidableLabs/victory/blob/5f60e9a700b9358a47792f3d137ef497abe71aee/test/jest/victory-line/victory-line.test.js#L162

However, this also might be a case for testing this functionality in a different place so we aren't using weird workarounds like this to access the props being passed into the individual components.

test/jest/victory-voronoi/victory-voronoi.test.js
```diff
@@ -0,0 +1,201 @@
...
+    });
+  });
+
+  describe("accessibility", () => {
+    it("adds an aria role to the path area", () => {
+      const { container } = render(<VictoryVoronoi />);
+      container.querySelectorAll("path").forEach((p) => {
+        const roleValue = p.getAttribute("role");
+        expect(roleValue).toEqual("presentation");
+      });
```
becca-bailey, line 178: Nicpick: We can replace this `forEach` with something like `expect(screen.getByRole('presentation')).toHaveLength(<expected path length>)`

becca-bailey (approved):
A couple questions and non-blockers, but otherwise looks good!

test/jest/victory-tooltip/victory-tooltip.test.js
scottrippey: I switched to `renderInSvg`, and removed these spies.  However, I still do get this warning; any ideas how to fix it?
```
  console.warn
    `renderInPortal` is not supported outside of `VictoryContainer`. Component will be rendered in place

      63 |         if (typeof this.context.portalUpdate !== "function") {
      64 |           var msg = "`renderInPortal` is not supported outside of `VictoryContainer`. " + "Component will be rendered in place";
    > 65 |           Log.warn(msg);
         |               ^
      66 |           this.renderInPlace = true;
      67 |         }
      68 |
```

test/jest/victory-tooltip/victory-tooltip.test.js
scottrippey: Sweet!  I added `jest-dom`, and this became:
```
expect(output).toBeInTheDocument();
expect(output).not.toBeVisible();
```

Also, this massively helped me find a mistake in this test too!  It should be `screen.getByText`, not the async `findByText` ... phew!

test/jest/victory-voronoi/victory-voronoi.test.js
scottrippey: I got the weird workaround to work, and was able to re-enable these skipped tests.  But I agree, we should consider improving this approach.

test/jest/victory-voronoi/victory-voronoi.test.js
scottrippey: I added `toHaveLength` assertions for all `querySelectorAll` queries
Surprisingly, one of the "labels" test was returning 0 labels.  This was also happening on the old test, so I didn't fix anything yet.  Maybe we need to fix this?

test/jest/victory-tooltip/victory-tooltip.test.js
scottrippey: Done!

test/jest/victory-voronoi/victory-voronoi.test.js
becca-bailey: Hmm, strange! Let's leave it for now with a comment to investigate more when we refactor.

PR #2253, opened 2022-05-23
Add tests for victory-stack, victory-group, and victory-selection-container

becca-bailey:
This is part of #2195.


heythisispaul (approved):
Looks good to me!

PR #2255, opened 2022-05-24
victory-histogram jest tests

heythisispaul:
Part of the effort for #2195, this PR migrates all existing `victory-histogram` tests to Jest and RTL.

### DIFF
--- a/test/jest/victory-histogram/victory-histogram.test.js
+++ b/test/jest/victory-histogram/victory-histogram.test.js
@@ -1,0 +1,211 @@
+import React from "react";
+import { render, screen, fireEvent } from "@testing-library/react";
+import { range } from "lodash";
+import { VictoryHistogram } from "victory-histogram";
+import { isBar, getBarHeight } from "../../svg-test-helper";
+
+describe("components/victory-histogram", () => {
+  const DATA_COMPONENT_ID = "data-component-id";
+  const DataComponent = () => <div data-testid={DATA_COMPONENT_ID} />;
+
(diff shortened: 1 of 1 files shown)


test/jest/victory-histogram/victory-histogram.test.js
```diff
@@ -0,0 +1,211 @@
...
+  const DATA_COMPONENT_ID = "data-component-id";
+  const DataComponent = () => <div data-testid={DATA_COMPONENT_ID} />;
+
+  describe("default component rendering", () => {
+    it("accepts user props", () => {
+      render(
+        <VictoryHistogram data-testid="victory-histogram" aria-label="Chart" />
+      );
+
+      const [svgNode] = screen.getAllByTestId("victory-histogram");
```
becca-bailey, line 17: I am running into this same issue with `victory-errorbar`, where these user props are being passed into both the container component and the group component, so there are two components with the same testid. If it's okay with you, I'll push up a change on this branch too once I fix it on mine so that we don't have to use `getAllByTestId` here.

becca-bailey (commented):
I'm going to wait to approve this until I push up a fix for the testid, but otherwise this looks good!

test/jest/victory-histogram/victory-histogram.test.js
heythisispaul: Alrighty, I rebased and this has been updated 👍

becca-bailey (approved):

PR #2256, opened 2022-05-25
Fix duplicate user props

becca-bailey:
This fixes a bug that Paul and I were running into where the user props like `data-testid` and `aria-label` were being passed in to both the container and the group component, so even when we were rendering a victory data component as a standalone component (without a container or VictoryChart wrapper), `screen.getByTestId` was returning multiple nodes in the tests rather than the parent wrapper component.

In order to fix this, I removed the user props logic from the `renderData` function in `add-events`, and passed the user props into the root component that was returned from a Victory component rather than the container. This ensures that the props will be passed into either the `svg` or the `g` component, depending on whether there is another container present. I also added some additional test coverage around this, which we should add to other component tests as we convert them.

### DIFF
--- a/packages/victory-core/src/victory-clip-container/victory-clip-container.js
+++ b/packages/victory-core/src/victory-clip-container/victory-clip-container.js
@@ -2,6 +2,7 @@ import React from "react";
 import PropTypes from "prop-types";
 import * as CustomPropTypes from "../victory-util/prop-types";
 import * as Helpers from "../victory-util/helpers";
+import * as UserProps from "../victory-util/user-props";
 import { assign, defaults, isObject, uniqueId } from "lodash";
 import ClipPath from "../victory-primitives/clip-path";
 import Circle from "../victory-primitives/circle";
@@ -84,6 +85,7 @@ export default class VictoryClipContainer extends React.Component {
   }
 
   renderClippedGroup(props, clipId) {
+    const userProps = UserProps.getSafeUserProps(props);
     const {
       style,
       events,
@@ -106,7 +108,7 @@ export default class VictoryClipContainer extends React.Component {
     );
     return React.cloneElement(
       groupComponent,
-      { ...groupProps, "aria-label": props["aria-label"], tabIndex },
+      { ...groupProps, tabIndex, ...userProps },
       [clipComponent, ...React.Children.toArray(children)]
     );
   }
--- a/test/jest/rendering-utils.js
+++ b/test/jest/rendering-utils.js
@@ -1,6 +1,5 @@
 import { render } from "@testing-library/react";
-import * as React from "react";
 
 export const renderInSvg = (component) => {
-  return render(<svg>{component}</svg>);
+  return render(component, { wrapper: "svg" });
 };
--- a/test/jest/victory-area/victory-area.test.js
+++ b/test/jest/victory-area/victory-area.test.js
@@ -1,17 +1,43 @@
+import "@testing-library/jest-dom";
 import { fireEvent, render, screen } from "@testing-library/react";
-import { curveCatmullRom } from "victory-vendor/d3-shape";
 import { range } from "lodash";
 import React from "react";
 import { Area, VictoryArea } from "victory-area";
+import { VictoryChart } from "victory-chart";
+import { curveCatmullRom } from "victory-vendor/d3-shape";
 import { calculateD3Path } from "../../svg-test-helper";
 
 describe("components/victory-area", () => {
   describe("default component rendering", () => {
-    it("accepts user props", () => {
-      render(<VictoryArea data-testid="victory-area" aria-label="Chart" />);
+    it("attaches safe user props to the container component", () => {
+      render(
+        <VictoryArea
+          data-testid="victory-area"
+          aria-label="Chart"
+          unsafe-prop="test"
+        />
+      );
 
-      expect(screen.queryByTestId("victory-area")).toBeDefined();
+      const container = screen.getByTestId("victory-area");
       expect(screen.getByLabelText("Chart")).toBeDefined();
+      expect(container).not.toHaveAttribute("unsafe-prop");
+      expect(container.nodeName).toEqual("svg");
+    });
+
+    it("attaches safe user props to the group component if the component is rendered inside a VictoryChart", () => {
+      render(
+        <VictoryArea
+          data-testid="victory-area"
+          aria-label="Chart"
+          unsafe-prop="test"
+        />,
+        { wrapper: VictoryChart }
+      );
+
+      const container = screen.getByTestId("victory-area");
+      expect(screen.getByLabelText("Chart")).toBeDefined();
+      expect(container).not.toHaveAttribute("unsafe-prop");
+      expect(container.nodeName).toEqual("g");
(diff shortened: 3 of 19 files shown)


packages/victory-core/src/victory-clip-container/victory-clip-container.js
```diff
@@ -84,6 +85,7 @@ export default class VictoryClipContainer extends React.Component {
   }
 
   renderClippedGroup(props, clipId) {
+    const userProps = UserProps.getSafeUserProps(props);
```
becca-bailey, line 88: I don't love that we have to explicitly call this here when we are also calling this function in the top-level component. However, there are props here that we can't pass into the base HTML component, so we do need some way of identifying which "extra" props are okay to pass in (using `...rest` creates DOM warnings). There is a bigger issue here, which is the issue where we are passing this props object around, and don't have visibility into what is in it.

test/jest/rendering-utils.js
```diff
@@ -1,6 +1,5 @@
 import { render } from "@testing-library/react";
-import * as React from "react";
 
 export const renderInSvg = (component) => {
-  return render(<svg>{component}</svg>);
+  return render(component, { wrapper: "svg" });
 };
```
becca-bailey, line 5: I had forgotten that there is a slightly easier way to add a custom wrapper component. Does it still make sense to have this helper, or is it just as clear if we just add `{ wrapper: "svg" }` to each individual test where we need it, rather than having to remember to use this helper?

test/jest/victory-area/victory-area.test.js
```diff
@@ -1,17 +1,43 @@
...
 import { Area, VictoryArea } from "victory-area";
+import { VictoryChart } from "victory-chart";
+import { curveCatmullRom } from "victory-vendor/d3-shape";
 import { calculateD3Path } from "../../svg-test-helper";
 
 describe("components/victory-area", () => {
   describe("default component rendering", () => {
-    it("accepts user props", () => {
-      render(<VictoryArea data-testid="victory-area" aria-label="Chart" />);
+    it("attaches safe user props to the container component", () => {
```
heythisispaul, line 12: Good thinking adding this check onto each of these components 👍

packages/victory-core/src/victory-clip-container/victory-clip-container.js
heythisispaul: Reading through this PR really opened by eyes to how much "magic" is happening in Victory to get (the appropriate) props applied to children.

This whitelisting of props (`safeUserProps`) seems to be flexible enough, but I agree it leads to a lot of manual checks and only checks for known cases of "unsafe" props.

Are there ideas to handle this differently in a future major version?

heythisispaul:
Other than maybe deleting that unneeded test util, this looks good to me!

test/jest/victory-area/victory-area.test.js
```diff
@@ -1,17 +1,43 @@
...
+        <VictoryArea
+          data-testid="victory-area"
+          aria-label="Chart"
+          unsafe-prop="test"
+        />
+      );
+
+      const container = screen.getByTestId("victory-area");
+      expect(screen.getByLabelText("Chart")).toBeDefined();
+      expect(container).not.toHaveAttribute("unsafe-prop");
```
scottrippey, line 23: Isn't `toHaveAttribute` part of `jest-dom`?  I'm confused how this is working :)

test/jest/victory-area/victory-area.test.js
```diff
@@ -1,17 +1,43 @@
...
+          unsafe-prop="test"
+        />,
+        { wrapper: VictoryChart }
+      );
 
-      expect(screen.queryByTestId("victory-area")).toBeDefined();
+      const container = screen.getByTestId("victory-area");
       expect(screen.getByLabelText("Chart")).toBeDefined();
+      expect(container).not.toHaveAttribute("unsafe-prop");
+      expect(container.nodeName).toEqual("g");
```
scottrippey, line 40: Should we test this?  This seems like an assertion on an implementation detail ... but then again, I guess you could just be validating that the attributes were added to the correct component

test/jest/victory-bar/victory-bar.test.js
```diff
@@ -3,14 +3,42 @@
 import React from "react";
 import { render, fireEvent, screen } from "@testing-library/react";
 import { range } from "lodash";
+import { VictoryChart } from "victory-chart";
 import { VictoryBar, Bar } from "victory-bar";
 import { isBar, getBarHeight } from "../../svg-test-helper";
+import "@testing-library/jest-dom";
```
scottrippey, line 9: I don't see this added to `devDependencies` yet ... maybe it's a "transient" dependency already?  In my PR, I also configured it globally using `setupFilesBeforeEnv`, maybe you can cherry-pick that config?

scottrippey (approved):
Great changes, glad to see an easy way to add these userProps.
Minor changes requested for using `jest-dom`, otherwise good!

test/jest/victory-bar/victory-bar.test.js
scottrippey: (you don't really have to cherry-pick, it's fine to just wait for my PR to merge)

packages/victory-core/src/victory-clip-container/victory-clip-container.js
becca-bailey: Yep, two ideas. First, I think we should take a bunch of the logic for providing prop defaults and getting derived values from the component's props and put it in a context wrapper or some sort of top-level state. That way, we have a single source of truth for accessing the original props, defaults, and derived values. Right now that's happening in many different places with functions like `getBaseProps`, `getCalculatedProps`, and all these individual render functions. 

Second, I want to add explicit types to these helper and be more explicit about what is being passed into them to make it all a little less mysterious.

test/jest/rendering-utils.js
becca-bailey: I think I will just remove this so we don't have to remember another import. FYI @scottrippey

test/jest/victory-area/victory-area.test.js
becca-bailey: https://github.com/FormidableLabs/victory/blob/fa406575ba917fb2b84e2c6b31a5201fbd255b5d/test/jest/victory-area/victory-area.test.js#L1

test/jest/victory-bar/victory-bar.test.js
becca-bailey: I'll just fix this once your PR gets merged.

packages/victory-core/src/victory-clip-container/victory-clip-container.js
heythisispaul: Top-level context was the first thing I was thinking of as well. It'd be easy enough to define an API for the props of the top level component and build a context value with that, or even maybe introduce something like a `VictoryProvider` or something where the user can do it themselves.

This is helpful, thanks!

heythisispaul (approved):

becca-bailey:
@heythisispaul you should be able to merge this into your PR now.

Issue #2063, opened 2022-01-21
VictoryScatter: externalEventMutations prop type invalid warning


ninni426:
### The Problem
While using VictoryScatter we are observing a warning in the console for prop-type being invalid when we pass a function in the callback.

'Warning: Failed prop type: VictoryScatter: prop type `externalEventMutations[0].callback` is invalid; it must be a function, usually from the `prop-types` package, but received `undefined`.'

This issue is observed once we have upgraded 'react-scripts' and 'eslint' packages. 

![image](https://user-images.githubusercontent.com/20175744/150524222-e12b5732-4288-486c-9902-e6a1ff5c6d90.png)

The code is developed taking https://formidable.com/open-source/victory/docs/common-props#externalEventsMutations document as the reference.

![image](https://user-images.githubusercontent.com/20175744/150535482-b1b24e7b-caee-43a0-9607-28c6c2c3fbe2.png)

Ref: https://stackoverflow.com/questions/36127842/it-must-be-a-function-usually-from-react-proptypes 

Due to audits, we do need to upgrade our packages, could you please assist in resolving this warning?

Current dependencies:

  "dependencies": {
    "@azure/msal-browser": "^2.21.0",
    "@azure/msal-react": "^1.2.0",
    "@microsoft/applicationinsights-react-js": "^3.2.2",
    "@microsoft/applicationinsights-web": "^2.7.2",
    "axios": "^0.21.4",
    "bootstrap": "^4.6.0",
    "env-cmd": "^10.1.0",
    "react": "^17.0.1",
    "react-bootstrap": "^1.6.0",
    "react-datepicker": "^3.6.0",
    "react-dom": "^17.0.1",
    "react-redux": "^7.2.2",
    "react-router-dom": "^5.2.0",
    "react-scripts": "^5.0.0",
    "react-select": "^4.3.1",
    "react-tooltip": "^4.2.15",
    "redux": "^4.0.5",
    "victory": "^36.2.0",
    "web-vitals": "^1.1.0"
  },
 "devDependencies": {
    "@testing-library/jest-dom": "^5.16.1",
    "@testing-library/react": "^12.1.2",
    "@wojtekmaj/enzyme-adapter-react-17": "^0.4.1",
    "enzyme": "^3.11.0",
    "eslint": "^8.7.0",
    "eslint-plugin-react": "^7.28.0"
  }

CC: @boygirl

becca-bailey:
[...]

PR #2257, opened 2022-05-25
[FIX] Replace references to non-existent PropTypes.function with PropTypes.func

gustaff-weldon:
Fixes https://github.com/FormidableLabs/victory/issues/2063

Several places in Victory reference non-existent `PropTypes.function` validator. This seem to be a common error, as exactly the same case is pointed out in `prop-types` codebase:
https://github.com/facebook/prop-types/blob/4de0644a10a554d0a556daa39f029369bc007ea5/checkPropTypes.js#L55-L57

This causes warning in console:
```
'Warning: Failed prop type: VictoryScatter: prop type externalEventMutations[0].callback is invalid; it must be a function, usually from the prop-types package, but received undefined.'
```

This PR replaces validators with `PropTypes.func` (https://github.com/facebook/prop-types#usage)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1244,7 +1244,7 @@ VictoryCore
   - `externalEventMutations` prop format:
   ````js
   externalEventMutations: PropTypes.arrayOf(PropTypes.shape({
-   callback: PropTypes.function,
+   callback: PropTypes.func,
    childName: PropTypes.oneOfType([
      PropTypes.string,
      PropTypes.array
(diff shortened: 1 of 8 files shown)


gustaff-weldon:
@becca-bailey it would be great if this bugfix can be released quickly, currently we would need to resort to running from fork

becca-bailey:
@gustaff-weldon thanks for looking into this, I'll get this merged and released ASAP.

becca-bailey (approved):

becca-bailey:
We don't have permission to run the docs and CI publish scripts from a fork, which is why some CI steps were failing. We're looking into fixing this in the future, but I don't anticipate any issues with this particular PR since it is only updating prop types.

gustaff-weldon:
@becca-bailey thanks for super quick turn-around on releasing this 🙇

PR #2260, opened 2022-05-25
Typescript Phase 1: Build configuration

scottrippey:
Adds all configurations needed to get this project working with TypeScript. 

# TODO:

- [x] Eslint
- [x] Jest
- [x] Babel (`build-es` and `build-lib`)
- [x] Webpack (`build-dist-dev` and `build-dist-min`)
- [x] ~Karma (do we need to do this?)~ (Does not need TS conversion / configuration)
- [x] ~Output consolidated `.d.ts` files per package~ Will do separately
- [x] Storybook
- [x] Demo server

# Testing
All of the following commands should run successfully:
```sh
nps build-package-libs
nps typecheck
nps lint
nps karma
nps jest
nps storybook
nps server.dev.ts
```

More importantly, some things should fail if broken.  
Add the following code to `packages/victory-core/src/victory-util/point-path-helpers.ts`, and you should get these 3 errors:
```ts
function foo(arg) {}
//^^^^^^            ESLint: Expected a function expression. (func-style)
//           ^^^    TS7006: Parameter 'arg' implicitly has an 'any' type.
//                ^ ESLint: Unexpected empty function 'foo'. (@typescript-eslint/no-empty-function)
```
(to show the errors, run `nps typecheck` and `nps lint.src`)


package-scripts.js
```diff
@@ -71,10 +71,17 @@ module.exports = {
...
+      default: npsUtils.series.nps("typecheck.core"),
+      base: "tsc --noEmit",
+      demo: "tsc -p ./demo/tsconfig.json --noEmit",
+      core: "lerna exec --scope victory-core -- nps typecheck.base"
+    },
     check: {
       ci: npsUtils.series.nps(
         "format.ci",
         "lint",
+        "typecheck",
```
ryan-roemer, line 84: Separately take a look at everything with `compile-ts` -- I think we want to integrate/reove that?

scottrippey: Oh, I missed that script, looking into it now.  Seems like it's the same as the `typecheck.base` script I just created.

From what I can tell, all our previous TypeScript configuration was being used for the `demo` folder, right?  I will try to make sure I test compiling that folder.

ryan-roemer: I think we both generate typedefs like: https://unpkg.com/browse/victory@36.4.1/es/index.d.ts and we have the TS demo

scottrippey:
Regarding the visual UI Tests:
The only visual diffs were due to the Storybook upgrade, which changed the body padding from `8px` to `1em`.  I pushed a change to override this back, and all visual tests passed.  So, I reverted that change, and approved all UI changes.

tsconfig.json
```diff
@@ -1,37 +1,69 @@
...
+    // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
+
+    /* Experimental Options */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+
+    /* Advanced Options */
+    "skipLibCheck": true,                     /* Skip type checking of declaration files. */
+    "forceConsistentCasingInFileNames": true  /* Disallow inconsistently-cased references to the same file. */
+  }
```
becca-bailey, line 68: Are there any strong feelings about whether or not we should remove all these comments? I'm okay with keeping them if it will help us to make changes to this config later, but it does make this config look a bit messy.

becca-bailey (approved):

tsconfig.json
scottrippey: I actually love the comments, it makes it super easy to see which options aren't being used, and to know which options are.  
Since I plan on modifying the config as we migrate, I think it'll be useful.

PR #2264, opened 2022-05-26
victory-vendor: script cleanup and exports enhancements

### LINKED ISSUE #2250: victory-vendor: misc improvements
Some things we could do:

- [x] Add `package.json:exports` for `package.json`. This is generally useful to have for npm packages using exports to introspect the project itself.
- [x] Add `package.json:exports` for `lib/*` consumers can point to CommonJS directly.
- [x] Add `package.json:exports` for `es/*` consumers can point to ESM directly.

ryan-roemer:
Adds the following to our `package.json:exports` for more flexible consumption:

- Add `package.json:exports` for `package.json`. This is generally useful to have for npm packages using exports to introspect the project itself.
- Add `package.json:exports` for `lib/*` consumers can point to CommonJS directly.
- Add `package.json:exports` for `es/*` consumers can point to ESM directly.

Additional things:

- Fix our `build.js` script to not errantly output an empty `victory-vendor/d3-*` directory.

Fixes #2250

### DIFF
--- a/packages/victory-vendor/package.json
+++ b/packages/victory-vendor/package.json
@@ -16,10 +16,13 @@
   "author": "Formidable",
   "license": "MIT AND ISC",
   "exports": {
+    "./package.json": "./package.json",
     "./d3-*": {
       "import": "./es/d3-*.js",
       "default": "./lib/d3-*.js"
-    }
+    },
+    "./es/d3-*": "./es/d3-*.js",
+    "./lib/d3-*": "./lib/d3-*.js"
   },
   "scripts": {
     "version": "yarn run build",
--- a/packages/victory-vendor/package.json
+++ b/packages/victory-vendor/package.json
@@ -20,9 +20,7 @@
     "./d3-*": {
       "import": "./es/d3-*.js",
       "default": "./lib/d3-*.js"
-    },
-    "./es/d3-*": "./es/d3-*.js",
-    "./lib/d3-*": "./lib/d3-*.js"
+    }
   },
   "scripts": {
     "version": "yarn run build",
--- a/packages/victory-vendor/scripts/build.js
+++ b/packages/victory-vendor/scripts/build.js
@@ -80,16 +80,14 @@ const main = async () => {
   const EsmBasePath = path.resolve(__dirname, `../es`);
   const CjsBasePath = path.resolve(__dirname, `../lib`);
   const VendorBasePath = path.resolve(__dirname, `../lib-vendor`);
+  const baseDirs = [EsmBasePath, CjsBasePath, VendorBasePath];
+  const cleanGlobs = [].concat(baseDirs, path.resolve(__dirname, "../d3-*"));
+
   log("Cleaning old vendor directories.");
+  await Promise.all(cleanGlobs.map((glob) => rimrafP(glob)));
+  log("Creating empty vendor directories.");
   await Promise.all(
-    [
-      EsmBasePath,
-      CjsBasePath,
-      VendorBasePath,
-      path.resolve(__dirname, "../d3-*")
-    ].map((libPath) =>
-      rimrafP(libPath).then(() => fs.mkdir(libPath, { recursive: true }))
-    )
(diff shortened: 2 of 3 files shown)


becca-bailey (approved):

packages/victory-vendor/scripts/build.js
```diff
@@ -80,16 +80,14 @@ const main = async () => {
   const EsmBasePath = path.resolve(__dirname, `../es`);
   const CjsBasePath = path.resolve(__dirname, `../lib`);
   const VendorBasePath = path.resolve(__dirname, `../lib-vendor`);
+  const baseDirs = [EsmBasePath, CjsBasePath, VendorBasePath];
+  const cleanGlobs = [].concat(baseDirs, path.resolve(__dirname, "../d3-*"));
```
scottrippey, line 84: I noticed that when running `nps clean.all` (at the root), it doesn't clean the `lib-vendor` folder (since it's just hard-coded to clean `es, lib` folders).  Would it be much work to include `lib-vendor` when cleaning all?

packages/victory-vendor/package.json
```diff
@@ -16,10 +16,13 @@
   "author": "Formidable",
   "license": "MIT AND ISC",
   "exports": {
+    "./package.json": "./package.json",
     "./d3-*": {
       "import": "./es/d3-*.js",
       "default": "./lib/d3-*.js"
-    }
+    },
+    "./es/d3-*": "./es/d3-*.js",
```
scottrippey, line 24: If I understand correctly, we're building + distributing the `d3-*` libraries because they ONLY support ESM, and Victory needs to support CommonJS, right?
But if the Victory consumer supports ESM, we'll pass-through the raw `d3-*` libraries as dependencies.

So my question is, why not just supply the bundled `d3-*` libraries for both, and remove `d3-*` as dependencies?  (as a way to reduce download size)

scottrippey (approved):
Thanks for fixing that `d3-*` folder!  I was pretty confused about its existence :)
Tested the builds locally, worked great.

packages/victory-vendor/package.json
```diff
@@ -16,10 +16,13 @@
...
+    "./lib/d3-*": "./lib/d3-*.js"
```
scottrippey, line 25: Another random question ...
Do Victory consumers import `victory-vendor` directly?  I was under the impression that this was just an internal module.
If it's just internal, then we probably don't need to have explicit `es` and `lib` export paths, since internally we don't use them, so these exports would only be available externally, and might result in 2 copies of the dependency being used.  Does that make sense?

ryan-roemer: I don't expect external consumers to likely need this -- I was more thinking of a way to be able to force the right library to use if the runtime choose something different, as I hit some issues with Jest being in one Mode when the execution environment was different.

But as we don't have that problem now, I can just remove the entries.

packages/victory-vendor/package.json
ryan-roemer: It will actually increase download size (for victory download size, would be a wash overall) to copy out all the ESM files and remove from `package.json:dependencies`. And for maintenance reasons, I would offer it's preferable to use ye olde `package.json:dependencies` to gate that.

packages/victory-vendor/scripts/build.js
ryan-roemer: Ah, I should add `--ignore victory-vendor` to the root `package-scripts.js`. We're off the main path here and our nps/lerna setup isn't great for easy overrides of just small things, so the build script is the right place to handle this for vendor.

ryan-roemer: And we also need to clean `victory-vendor/d3-*` files too, so I don't think we want to shove all that in root task.

ryan-roemer:
@becca-bailey thanks for the review!

@scottrippey thanks for the review and comments! I think https://github.com/FormidableLabs/victory/pull/2264/commits/537219d6763b92e46d0331cab3199dc1b6222519 takes care of everything for you?

scottrippey (approved):
Looks great! Thanks for the answers.

PR #2266, opened 2022-05-27
Convert tests for victory error bar

becca-bailey:
This is part of #2195 

This also includes removing the `renderInSvg` function and using `{ wrapper: 'svg' }` instead. I meant to push up this change as a separate PR, but it ended up here instead.

### DIFF
--- a/packages/victory-errorbar/src/error-bar.js
+++ b/packages/victory-errorbar/src/error-bar.js
@@ -1,7 +1,7 @@
 /* eslint-disable max-statements */
 import React from "react";
 import PropTypes from "prop-types";
-import { Helpers, CommonProps, Line } from "victory-core";
+import { Helpers, CommonProps, Line, UserProps } from "victory-core";
 import { assign } from "lodash";
 
 const renderBorder = (props, error, type) => {
@@ -17,7 +17,8 @@ const renderBorder = (props, error, type) => {
     x1: vertical ? error[type] : props.x - props.borderWidth,
     x2: vertical ? error[type] : props.x + props.borderWidth,
     y1: vertical ? props.y - props.borderWidth : error[type],
-    y2: vertical ? props.y + props.borderWidth : error[type]
+    y2: vertical ? props.y + props.borderWidth : error[type],
+    "data-type": `border-${type}`
   });
 };
 
@@ -34,7 +35,8 @@ const renderCross = (props, error, type) => {
     x1: props.x,
     x2: vertical ? props.x : error[type],
     y1: props.y,
-    y2: vertical ? error[type] : props.y
+    y2: vertical ? error[type] : props.y,
+    "data-type": `cross-${type}`
   });
 };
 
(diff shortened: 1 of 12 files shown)


packages/victory-errorbar/src/error-bar.js
```diff
@@ -17,7 +17,8 @@ const renderBorder = (props, error, type) => {
     x1: vertical ? error[type] : props.x - props.borderWidth,
     x2: vertical ? error[type] : props.x + props.borderWidth,
     y1: vertical ? props.y - props.borderWidth : error[type],
-    y2: vertical ? props.y + props.borderWidth : error[type]
+    y2: vertical ? props.y + props.borderWidth : error[type],
+    "data-type": `border-${type}`
```
becca-bailey, line 21: I don't usually like adding arbitrary props like this, but previously we were getting this type by string matching the key, and that felt wrong to me.

test/jest/victory-errorbars/error-bar.test.js
```diff
@@ -0,0 +1,108 @@
...
+      x: d3Scale.scaleLinear(),
+      y: d3Scale.scaleLinear()
+    },
+    borderWidth: 20
+  };
+
+  const compareLineCoordinates = (line, coordinates) => {
+    forEach(coordinates, (coordinateValue, coordinateName) => {
+      expect(parseFloat(line.getAttribute(coordinateName), 10)).toEqual(
+        parseFloat(coordinateValue, 10)
```
scottrippey, line 23: `parseFloat` does not accept a radix arg, so it's not necessary here.
```suggestion
      expect(parseFloat(line.getAttribute(coordinateName))).toEqual(
        parseFloat(coordinateValue)
```

test/jest/victory-errorbars/victory-errorbars.test.js
```diff
@@ -0,0 +1,1087 @@
...
+          const yScale = d3Scale
+            .scaleLinear()
+            .domain([-0.2, 5.2])
+            .range([
+              svgDimensions.height - svgDimensions.padding,
+              svgDimensions.padding
+            ]);
+
+          const bars = screen.getAllByTestId("error-bar");
+          bars.forEach((node, i) => {
```
scottrippey, line 333: We should probably always use a `.toHaveLength(...)` assertion, so that we're sure the following assertions are running.  Or maybe `.not.toHaveLength(0)` if the quantity doesn't matter.

scottrippey (approved):
Looks good.  Minor comments.

test/jest/victory-errorbars/error-bar.test.js
```diff
@@ -0,0 +1,108 @@
...
+      { x1: 4, x2: 1, y1: 5, y2: 5 },
+      // Left Cross(x, negativeErrorX, y, y)
+      { x1: 4, x2: 3, y1: 5, y2: 5 },
+      // Bottom Cross(x, x, y, negativeErrorY)
+      { x1: 4, x2: 4, y1: 5, y2: 0.2 },
+      // Bottom Cross(x, x, y, positiveErrorY)
+      { x1: 4, x2: 4, y1: 5, y2: 2 }
+    ];
+
+    expect(lines.length).toEqual(8);
```
heythisispaul, line 53: This should probably be switched to a `toHaveLength()` assertion for uniformity.

test/jest/victory-errorbars/victory-errorbars.test.js
heythisispaul: Agreed, or an `expect.hasAssertions()` to verify that the `forEach` has at least one element.

heythisispaul (approved):
Just a few opinions on changing some assertion syntax, but this looks good to me!

PR #2271, opened 2022-05-30
Typescript: Generate proper .d.ts files for Core

scottrippey:
# What
- Adds a "types" script which generates the `.d.ts` output for our builds.
- Adds a `tsconfig.json` for each package (which just extends the base config).  This is necessary to set the proper "scope" for each build ... otherwise TypeScript will always compile _everything_.

# Migration Pattern

Currently, each package has a top-level `index.d.ts` file, which contains type definitions for all exported members.
This is problematic, because when a method signature changes, the type definition must be updated manually.

The best, but hardest, solution is to convert all `.js` files to `.ts`.  While we do anticipate this being the "end game" for us, it's a lot of work to handle at once. 

Instead, here's a way to migrate in steps:

1. Split the `index.d.ts` contents into separate files.  
  For example, I split the `victory-core/src/index.d.ts` file into a bunch of files, like `victory-accessible-group.d.ts`, `victory-animation.d.ts`, `victory-label.d.ts` and so on.  I moved those files into the matching folder, alongside the matching `.js` file.
2. Delete the empty `index.d.ts` file
3. Rename `index.js` to `index.ts`.  This usually doesn't require any changes.

That's it!  When we create our builds, all our `.d.ts` files will be copied to the output.  Also, all `.ts` files will be compiled, and the generated `.d.ts` files will be in the output as well; especially the `index.d.ts`.

# Testing

Build everything via:
```sh
yarn nps build-package-libs
```

- Verify the output of the folders has proper `.d.ts` files, like:
image

- Verify that the top `index.d.ts` should be the same as the `index.ts`, just a bunch of exports, like:
image

- TODO: Somehow verify that the "published types" are working from a "consumer" perspective?


becca-bailey (approved):
Looks good to me - these type files are much more manageable when we move them into their respective directories. Do you anticipate any problems with us releasing this as a patch version bump? In other words, moving these types around won't change anything for users importing types from Victory, correct?

scottrippey:
The main difference right now is that we're not exporting * yet. So the props are not yet being exported. I've fixed this already in the next PR

PR #2274, opened 2022-05-31
TS migration: victory-primitives

scottrippey:
Migrates all victory-primitives to TypeScript


packages/victory-core/src/victory-label/victory-label.tsx
```diff
@@ -637,6 +691,7 @@ VictoryLabel.propTypes = {
     PropTypes.func,
     PropTypes.array
   ]),
+  // @ts-expect-error Function is not assignable to string"
```
becca-bailey, line 694: Are all these TS errors conflicts in the prop types definitions themselves, or is TS doing some type inference using these prop types that we're trying to disable?

scottrippey: In this case, TS knows that `textAnchor` is defined as a `string`, so it's complaining that the `propTypes` here don't match, since they allow a function to be passed.
I would like to have a chat soon to discuss how to better handle these "props that can be functions", hopefully in a generic way.

becca-bailey: In my prototype, I have a `ValueOrAccessor<ValueType, PropsType = {}>` type that I used for any prop value that can be passed in as a value or an accessor. Should we consider something similar here?

PR #2275, opened 2022-05-31
Disable func-style eslint rule

becca-bailey:
We are disabling this eslint rule in many places, and personally I don't think enforcing `const` over `function` syntax has a lot of value. Any objections to disabling it and moving on with our lives?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,6 +15,7 @@ module.exports = {
     "plugin:react-hooks/recommended"
   ],
   rules: {
+    "func-style": "off",
     "react/sort-comp": "off",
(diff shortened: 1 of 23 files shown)


.eslintrc.js
```diff
@@ -15,6 +15,7 @@ module.exports = {
     "plugin:react-hooks/recommended"
   ],
   rules: {
+    "func-style": "off",
```
scottrippey, line 18: Agreed! https://github.com/FormidableLabs/victory/pull/2274/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R18

scottrippey (approved):
LGTM!

PR #2277, opened 2022-06-01
Enable `eslint-comments`

scottrippey:
This plugin enforces some rules for `// eslint-disable` style comments.  
Mainly, this requires us to clean-up any comments that are no longer necessary!


becca-bailey (approved):
🎉

PR #2278, opened 2022-06-01
ts-migration: migrated core components to TS

scottrippey:


package.json
```diff
@@ -141,5 +141,8 @@
     "webpack-dev-server": "^4.9.0",
     "webpack-stats-plugin": "^1.0.3"
   },
+  "resolutions": {
+    "@types/react": "^16.14.26"
```
scottrippey, line 145: This little gem tells yarn to only install 1 copy of these types.  Without it, there was a transient-dependency that was including `@types/react: 18` and was messing up typechecks in various places.

packages/victory-core/src/victory-animation/victory-animation.tsx
```diff
@@ -163,21 +236,21 @@ export default class VictoryAnimation extends React.Component {
         setTimeout(() => {
           this.loopID = this.timer.subscribe(
             this.functionToBeRunEachFrame,
-            this.props.duration
+            this.props.duration!
```
scottrippey, line 239: We have a general problem with our types.  Since we use `defaultProps`, some of our types are optional to the "outside world" but are always defined internally.

Typically, I solve this in FC's by using defaults when deconstructing.  Not sure if there's a better way to handle that for Class Components.

package.json
```diff
@@ -125,7 +125,7 @@
     "nps-utils": "^1.5.0",
     "prettier": "^2.6.2",
     "prop-types": "^15.8.1",
-    "react": "^16.13.1",
+    "react": "^16.14.0",
```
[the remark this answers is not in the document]
scottrippey: Yep that solves the problem too!  I can remove the "resolutions" bit

becca-bailey (approved):

PR #2285, opened 2022-06-03
TypeScript: fix a few build issues

scottrippey:
# Fixes:
- Use named import for `addEvents > Transition` dependency
- Properly type the `context` property for class components, so that it doesn't get initialized
- Adds "optional chaining" transpilation to Babel

# Testing
These issues were all found in React Native, although it's not specific to that.
To test:
```
cd demo/rn
yarn start
```
Open in iOS simulator, and verify that all demos are working as expected.

image

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -13,6 +13,7 @@ module.exports = {
     "@babel/transform-spread",
     "@babel/transform-template-literals",
     "@babel/proposal-object-rest-spread",
+    "@babel/proposal-optional-chaining",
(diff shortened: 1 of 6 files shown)


scottrippey:
2 things worth noting …
1. If we switched to `babel-env` it might’ve been easier to catch the “optional chaining” support
2. We still need to convert `victory-utils` to TS, which would catch the named import bug

becca-bailey (approved):

PR #2288, opened 2022-06-04
[Demo]: Upgrade React Native demo app

jpdriver:
- @scottrippey recently reported that the RN Demo no longer runs on physical devices
- Expo regularly drop older SDK versions from their Expo Go app
- this PR bumps package versions but leaves the demo `/src` untouched

|Package|Before|After|
|-|-|-|
|`expo`|`~42.0.1`|`~45.0.0`|
|`react`|`16.13.1`|`17.0.2`|
|`react-native`|`https://github.com/expo/react-native/archive/sdk-42.0.0.tar.gz`|`0.68.2`|

## Screenshots

|Before|After|
|-|-|
|IMG_3188|IMG_3189|


scottrippey (approved):
Looks good, thanks!

demo/rn/App.tsx
```diff
@@ -1,7 +1,10 @@
 import React from "react";
+import { LogBox } from "react-native";
 import { NavigationContainer } from "@react-navigation/native";
 import { RootNavigator } from "./src/screens/root-navigator";
 
+LogBox.ignoreLogs(["Require cycle: ../../packages/victory"]);
```
gksander, line 6: Are we getting require cycles out of the box with victory-native? Or is this an artifact of the demo setup? (Just want to make sure end-users aren't suffering through require cycles, and if they are – put something on our roadmap to address that.)

gksander (approved):
Nice 👍

demo/rn/App.tsx
jpdriver: they're caused by upstream issues with `d3` that we can't easily address 😓 

more info and guidance [here](https://formidable.com/open-source/victory/docs/native#4-ignoring-require-cycles)

PR #2290, opened 2022-06-07
Modularize calculated props helpers

becca-bailey:
This adds custom hooks that can be used to calculate Victory props inside a context provider component.

`useDomain` - Accepts domain props and returns a domain for the data formatted as `[min, max]` (currently excludes domain padding logic)
`useRange` - Returns a range using height, width, and padding
`useScale` - Returns the base scale (no domain or range) from a string or scale prop
`useData` - Returns normalized data 

This also includes a basic `VictoryProvider` component that stores these calculated values, and a demo component that uses this provider. 

The Victory bar demo component demonstrates how we can use these values for rendering data, but it does not include events, styling, or animations. 

To test:
- Change the export in `victory-bar/index.js` from `victory-bar` to `victory-bar-beta`. 
- Skip the `describe` block in `victory-bar.test.js` that contains events and run `yarn nps "jest victory-bar". This component should pass all other tests for victory bar.


packages/victory-core/src/victory-util/hooks/use-domain.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function getValueForAxis<T = unknown>(
+  value: T | ForAxes<T> | undefined,
+  axis: Axis
+): T | undefined {
+  if (hasValueForAxis<T>(value, axis)) {
+    return value[axis] as T;
+  }
+  return value;
+}
```
becca-bailey, line 55: These will probably end up being shared helpers, since I see this logic repeated in a lot of places. Many victory props can be passed in as either `T` or `{ x: T, y: T }`, and these helpers provide some type safety for these props.

packages/victory-core/src/victory-util/hooks/use-domain.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function getDomainFromMinMax(min: Datum, max: Datum): Tuple<Datum> {
+  // TODO: Victoy currently has some really specific logic in getDomainFromMinMax
+  // that adds or subtracts a very small number from each domain to avoid the min
+  // and max being the same value. This has resulted in some weird behavior in the
+  // past, so we should revisit this.
```
becca-bailey, line 61: Right now, Victory has some logic so that if the min and max are the same value, rather than return the same value it will add or subtract a small amount so they are different values. For example, [1, 1] will become something [0.99, 1.01]. This can be confusing for users, as documented in issues like this one. https://github.com/FormidableLabs/victory/issues/1471

packages/victory-core/src/victory-util/hooks/use-domain.ts
```diff
@@ -0,0 +1,105 @@
...
+  return [min, max];
+}
+
+export function useDomain(
+  { data = [], ...props }: DomainProps,
+  axis: Axis,
+  includeZero = false
+): DomainTuple {
+  const allValues = React.useMemo(() => {
```
becca-bailey, line 70: I'm trying not to prematurely optimize, I was considering using a `getAllValues` callback function here instead since this could be a more performance-heavy operation, and it's only necessary if there are not domain props. (Which is the most common use case for Victory users).

becca-bailey:
I changed the base so we can merge this to a feature branch for the initial state management work.

packages/victory-core/src/victory-util/hooks/use-domain.ts
```diff
@@ -0,0 +1,100 @@
...
+import { DomainTuple, DomainValue } from "../types";
+
+interface DomainProps {
+  domain?: ValueOrAxesObject<DomainTuple>;
+  maxDomain?: ValueOrAxesObject<number>;
+  minDomain?: ValueOrAxesObject<number>;
+  domainPadding?: ValueOrAxesObject<number | Tuple<number>>;
+  width?: number;
+  height?: number;
+  padding?: PaddingProps;
```
becca-bailey, line 14: In a perfect world there would be one component props interface I could pick these props from, but they seem to be scattered around a few different type definitions. The structure of these types might be worth re-evaluating, even if it ends up being a breaking change.

scottrippey:
So far, this  is looking great.  I really like that "consuming" these computed props will be very simple.

The only thing I wanted to mention, is that there's a lot of `useMemo` which really makes it harder to read.  Maybe we can avoid `useMemo` for now, and address any possible performance issues later?

packages/victory-bar/src/victory-bar-beta.tsx
```diff
@@ -0,0 +1,123 @@
...
+    { x: 1, y: 1 },
+    { x: 2, y: 2 },
+    { x: 3, y: 3 },
+    { x: 4, y: 4 }
+  ],
+  height: 300,
+  includeZero: true,
+  padding: 50,
+  // TODO: Why do I have to cast this?
+  sortOrder: "ascending" as "ascending" | "descending",
```
becca-bailey, line 31: Why do I have to cast this, even when I am explicitly typing the props?

packages/victory-core/src/types/prop-types.ts
```diff
@@ -62,37 +88,21 @@ export interface EventPropTypeInterface<TTarget, TEventKey> {
...
+//   (input: NumberValue): number;
+
+//   ticks: (count?: number) => number[];
+//   tickFormat: (count?: number) => (d: number) => string;
+//   domain: () => number[];
+//   range: () => TRange[];
+//   copy: () => this;
+//   invert: (value: number) => number;
+// }
+export type D3Scale = any;
```
becca-bailey, line 105: I will keep looking into this but I have spent many hours trying to type this in a way that works for d3 scale types, and I haven't figured it out yet.

packages/victory-bar/src/victory-bar-beta.tsx
```diff
@@ -0,0 +1,122 @@
...
+        bottomLeft?: NumberOrCallback;
+        bottomRight?: NumberOrCallback;
+      };
+  dataComponent?: React.ReactElement;
+  groupComponent?: React.ReactElement;
+  horizontal?: boolean;
+}
+
+// TODO: This would be a shared helper that allows us to access context values in the base component
+export function withContainer<Props extends VictoryCommonProps>(
```
scottrippey, line 57: I like this wrapper!

packages/victory-bar/src/victory-bar-beta.tsx
```diff
@@ -0,0 +1,122 @@
...
+    return (
+      <VictoryProvider {...providerProps}>
+        <WrappedComponent {...props} />
+      </VictoryProvider>
+    );
+  };
+}
+
+function VictoryBar({
+  dataComponent = <Bar />,
```
scottrippey, line 88: Gotta love default props! So much easier than `defaultProps` :)

packages/victory-bar/src/victory-bar-beta.tsx
```diff
@@ -0,0 +1,122 @@
...
+  groupComponent = <g role="presentation" />,
+  alignment,
+  barRatio,
+  barWidth,
+  cornerRadius,
+  horizontal = false
+}: VictoryBarProps) {
+  const { scale, data, domain } = useVictoryContext();
```
scottrippey, line 96: ❤️

packages/victory-core/src/victory-state/use-data.ts
```diff
@@ -0,0 +1,88 @@
...
+  sortKey,
+  sortOrder = "ascending",
+  ...props
+}: DataProps) {
+  const data = React.useMemo<Datum[]>(() => {
+    if (props.data) {
+      return props.data;
+    }
+    return [];
+  }, [props.data]);
```
scottrippey, line 48: `useMemo` not necessary
```suggestion
  const data = props.data || [];
```

scottrippey (commented):
Still reviewing but taking a break :)

packages/victory-core/src/victory-state/use-scale.ts
```diff
@@ -0,0 +1,67 @@
...
+    case "log":
+      return scaleLog;
+    case "sqrt":
+      return scaleSqrt;
+    default:
+      return DEFAULT_SCALE;
+  }
+}
+
+export function useScale(
```
scottrippey, line 44: Ok, I have a topic to discuss ...
I'm thinking about these new hooks -- `useScale` `useData` `useDomain` and `useRange` ... 

I would expect to use these hooks in my own components.  Like
```
const domain = useDomain();
const scale = useScale();
const range = useRange();
```
And I would kinda expect the values to be coming from the Context.

The way these hooks are written, currently, is to be used by the `VictoryProvider`.  And one thing I notice:  they might not need to be hooks.  These methods don't really use `useState` or `useEffect` or `useContext` ... they primarily just use `useMemo`.

So here's what I'd like to think about -- maybe I'm wrong about my understanding, so please let me know.
If these methods were changed to something like `normalizeScale`, `normalizeDomain`, etc ..., and we removed the `useMemo` out of them, would that make sense?
We would move the `useMemo` into the `VictoryProvider` itself, so that we're memoizing the results of `normalizeScale` etc.

And, then we could export some of these "consumer hooks" like:
```
export function useScale() {
  return useContextSelector(VictoryContext, c => c.scale);
}
```

Let me know if my explanation makes sense!

scottrippey (commented):
Second half of the review :)

packages/victory-core/src/victory-state/victory-provider.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+type ContextValue = ContextType | null;
+
+export function useVictoryContext<T>(selector: (value: ContextValue) => T): T {
+  const context = useContextSelector<ContextValue, ContextValue>(
+    VictoryContext,
+    (v) => v
+  );
+
+  if (!context) {
```
becca-bailey, line 79: For now, I'm just null checking the entire context value before trying to extract values from it.

packages/victory-core/src/victory-state/victory-provider.tsx
```diff
@@ -0,0 +1,96 @@
...
+    throw new Error("useVictoryContext must be used within a VictoryProvider");
+  }
+
+  return useContextSelector<ContextValue, T>(VictoryContext, selector);
+}
+
+export function useScale() {
+  return useVictoryContext<ScaleType>((value) => value!.scale);
```
becca-bailey, line 87: It's a little annoying that we're null-checking the context value, and also asserting that it is non-null when using the value. There should be a better way to type this...

packages/victory-core/src/victory-state/victory-provider.tsx
```diff
@@ -21,27 +24,38 @@ export function VictoryProvider({
   ...props
 }: VictoryProviderProps) {
   // TODO: Get data
-  const xDomain = useDomain(props, "x", includeZero);
-  const yDomain = useDomain(props, "y", includeZero);
-  const domain = { x: xDomain, y: yDomain };
+  const domain = React.useMemo(
```
scottrippey, line 27: Love the new improvements, looks great!

scottrippey (approved):

packages/victory-core/src/victory-state/victory-provider.tsx
scottrippey: It's good enough for now.  Maybe this would do it?
```suggestion
  return useVictoryContext<ScaleType>((value) => value!.scale!);
```

becca-bailey: I think just the null check on the `value` is sufficient. The reason I kept the default value of the context `null` rather than something like `Symbol('no context')` is because I would have needed to check the type of the context every time, and at least a non-null assertion like `!` is easier than a typeof conditional in every single selector.

becca-bailey: Nevermind, I think I figured it out. I just needed to make sure the value is typed to be non-nullable in the selector type signature.

PR #2299, opened 2022-06-10
TypeScript migration: more utils!

scottrippey:
# What
Migrates more utils to TypeScript, of course!

After this PR, there are only 3 more `js` files left to convert (`add-events`, `events`, and `prop-types`).


packages/victory-core/src/victory-util/helpers.ts
```diff
@@ -218,13 +222,24 @@ export function getCurrentAxis(axis, horizontal) {
  * @returns {Array} returns an array of results from calling the iteratee on all nested children
  */
 /* eslint-disable max-params */
-export function reduceChildren(
-  children,
-  iteratee,
+export function reduceChildren<
```
becca-bailey, line 225: This function is also on my list of really expensive functions to get rid of at some point, so it's fine if it's missing some types.

becca-bailey (approved):
LGTM!

PR #2300, opened 2022-06-13
TS migrate: prop types

scottrippey:
Migrates our custom PropType validators to TS.

Some of our types are now marked as `isRequired` to match the TypeScript definition.


gksander (approved):
Nice.

PR #2301, opened 2022-06-14
Move tests to root directories

becca-bailey:
This is the initial work to move test files into the packages they are testing. I ran into some complications around build config, detailed in comments below.

PR #2305, opened 2022-06-16
Build enhancements for move tests PR

ryan-roemer:
- Separates out `.d.ts` copying from `babel` to a dedicated copy task (using `cpx2`)
- Adds watch support for babel, tsc, and copying

If we like this I think goal is merging commit back to #2301

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -75,9 +75,15 @@ module.exports = {
       packages: "lerna exec --ignore victory-vendor -- nps typecheck.src"
     },
     types: {
-      base: "tsc -p ./tsconfig.build.json --emitDeclarationOnly --rootDir src",
-      lib: "nps types.base -- -- --outDir lib",
-      es: "nps types.base -- -- --outDir es"
+      create:
+        "tsc -p ./tsconfig.build.json --emitDeclarationOnly --rootDir src",
+      "create-lib": 'nps "types.create --outDir lib"',
+      "create-es": 'nps "types.create --outDir es"',
+      copy: "cpx 'src/**/*.d.ts'",
+      "copy-lib": 'nps "types.copy lib"',
+      "copy-es": 'nps "types.copy es"',
+      lib: npsUtils.concurrent.nps("types.create-lib", "types.copy-lib"),
+      es: npsUtils.concurrent.nps("types.create-es", "types.copy-es")
     },
     check: {
       ci: npsUtils.series.nps(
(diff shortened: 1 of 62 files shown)


package-scripts.js
```diff
@@ -75,9 +75,15 @@ module.exports = {
...
-      es: "nps types.base -- -- --outDir es"
+      create:
+        "tsc -p ./tsconfig.build.json --emitDeclarationOnly --rootDir src",
+      "create-lib": 'nps "types.create --outDir lib"',
+      "create-es": 'nps "types.create --outDir es"',
+      copy: "cpx 'src/**/*.d.ts'",
+      "copy-lib": 'nps "types.copy lib"',
+      "copy-es": 'nps "types.copy es"',
+      lib: npsUtils.concurrent.nps("types.create-lib", "types.copy-lib"),
+      es: npsUtils.concurrent.nps("types.create-es", "types.copy-es")
```
scottrippey, line 86: Nice job keeping this organized and reasonaboutable

package-scripts.js
```diff
@@ -95,8 +101,30 @@ module.exports = {
       default: npsUtils.series.nps("lint", "test")
     },
     watch: {
-      es: "lerna exec --parallel --ignore victory-native --ignore victory-vendor -- cross-env BABEL_ENV=es babel src --out-dir es --config-file ../../.babelrc.build.js --copy-files --extensions .tsx,.ts,.jsx,.js --watch",
-      lib: "lerna exec --parallel --ignore victory-native --ignore victory-vendor -- cross-env BABEL_ENV=commonjs babel src --out-dir lib --config-file ../../.babelrc.build.js --copy-files --extensions .tsx,.ts,.jsx,.js --watch",
+      debug:
```
scottrippey, line 104: Probably should remove this

package-scripts.js
```diff
@@ -95,8 +101,30 @@ module.exports = {
...
+        'lerna exec --parallel --ignore victory-native --ignore victory-vendor -- nps \\"types.copy es --watch\\"',
+      "babel-es":
+        "lerna exec --parallel --ignore victory-native --ignore victory-vendor -- cross-env BABEL_ENV=es babel src --out-dir es --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js --watch",
+      "babel-lib":
+        "lerna exec --parallel --ignore victory-native --ignore victory-vendor -- cross-env BABEL_ENV=commonjs babel src --out-dir lib --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js --watch",
+      lib: npsUtils.concurrent.nps(
+        "watch.babel-lib",
+        "watch.types-create-lib",
+        "watch.types-copy-lib"
+      ),
```
scottrippey, line 122: This seems to be working fine on my machine.  It takes a while to get started, but after a minute the CPU goes back to normal while waiting for changes.

However, I was going to make a suggestion.  It might be nice to reorganize these scripts a little bit, so that I can run "watch" for just 1 specific patckage (eg. `victory-core`)
For example, I would like to run:
```
lerna exec --scope victory-core "nps watch.lib"
```
and then maybe we could have a single `watch.default` which runs `lerna exec --parallel ...`

What do you think of this approach?

scottrippey (approved):
Everything works well and looks good!  
Just one minor comment on the `watch` scripts below.

package-scripts.js
scottrippey: Done

package-scripts.js
scottrippey: Done :) It's very easy now to run for specific packages!

PR #2306, opened 2022-06-17
TS migrate: add events

scottrippey:
# What
- Migrates the final part of `victory-utils`: `add-events`!
- With this conversion, we can now migrate other Victory components, like `victory-line`.


packages/victory-core/src/victory-util/add-events.tsx
```diff
@@ -288,24 +361,28 @@ export default (WrappedComponent, options) => {
...
 
-    animateComponent(props, defaultAnimationWhitelist) {
+    animateComponent(props: TProps, defaultAnimationWhitelist: string[]) {
       const animationWhitelist =
-        props.animate && props.animate.animationWhitelist
+        typeof props.animate === "object" &&
+        props.animate &&
+        props.animate.animationWhitelist
           ? props.animate.animationWhitelist
           : defaultAnimationWhitelist;
```
gksander, line 370: Might consider `props.animate?.animationWhitelist || defaultAnimationWhitelist`

packages/victory-core/src/victory-util/add-events.tsx
```diff
@@ -25,34 +28,104 @@ const datumHasXandY = (datum) => {
...
-    }
 
-    shouldComponentUpdate(nextProps) {
+      this.cacheValues(this.getCalculatedValues(props));
+    }
+    state = {};
+    getEventState = Events.getEventState.bind(this);
+    getScopedEvents = Events.getScopedEvents.bind(this);
+    getEvents = (p, target, eventKey) => {
+      return Events.getEvents.call(
```
gksander, line 112: `.call`, now that's a method I haven't seen in awhile 🤩

gksander (approved):
Nice!

packages/victory-core/src/victory-util/add-events.tsx
scottrippey: I know ... as I was migrating these `Events.*` methods, I was surprised to use the `this` argument typing again.  Not very common nowadays!

packages/victory-core/src/victory-util/add-events.tsx
scottrippey: Done

PR #2313, opened 2022-06-20
TypeScript: Mark `data` as `readonly`

### LINKED ISSUE #2309: Typescript readonly data array is incompatible with data property interface definition
### The Problem

I'm attempting to use a functional programming style to the extent it is possible to do so using the JavaScript frameworks I am using.

As part of this effort, I am defining my arguments and data constructs as readonly using TypeScript so that my data is not mutable.

However, I am experiencing an issue with Victory in that the TypeScript interface for the `data` property (on the `VictoryLine` component for example) cannot accept a readonly value because its typescript definition is using the mutable `any` type.

I get the following error:

`The type 'ChartData' is 'readonly' and cannot be assigned to the mutable type 'any[]'.`

### Reproduction

https://codesandbox.io/s/winter-tree-00wu4z?file=/index.tsx

### Potential Solution

In limited testing I tried changing the definition for the `data` property from `any[]` to `any[] | readonly any[]` and this solved the TypeScript error, but I haven't tested this properly.

If this issue is accepted, and the above is deemed an acceptable solution, then I will create a PR that modifies the TypeScript interfaces in the correct places.

scottrippey:
# What
Marks `data` as `readonly`.
We don't mutate the data, so this doesn't break anything, and doesn't require any other changes.

Fixes #2309

### DIFF
--- a/packages/victory-core/src/victory-util/common-props.tsx
+++ b/packages/victory-core/src/victory-util/common-props.tsx
@@ -21,7 +21,7 @@ import { NumberOrCallback, StringOrCallback } from "../types/callbacks";
 
 export interface VictoryDatableProps {
   categories?: CategoryPropType;
-  data?: any[];
+  data?: readonly any[];
   dataComponent?: React.ReactElement;
   domain?: DomainPropType;
   domainPadding?: DomainPaddingPropType;


gksander (approved):
Nice.

becca-bailey (approved):

PR #2315, opened 2022-06-20
Update prettier trailing comma rule

becca-bailey:
Updates prettier trailing commas from "none" to "all" (all includes TS types as well as objects) and runs `yarn nps format`.


scottrippey (approved):
That's a lotta commas!

scottrippey:
Maybe we should wait to merge this until we've finished #2301 to make it easier to fix conflicts?  Maybe it doesn't matter, I dunno.

PR #2316, opened 2022-06-20
Add Victory Bar v37

becca-bailey:
This functionality will work for rendering either one standalone `VictoryBar` component or a single nested `VictoryBar` component nested under a `VictoryChart`. 

As discussed, we will need to add functionality for two-way syncing between child props and the provider props so the provider knows about the props of each child component and we can support multiple datasets in child components.

Right now all new components live inside the `v37` directories, so this should not touch any existing behavior.


packages/victory-bar/src/v37/victory-bar.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const {
+    alignment,
+    barRatio,
+    barWidth,
+    dataComponent,
+    cornerRadius,
+    horizontal,
+    groupComponent,
+    data,
+  } = useVictoryProps<VictoryBarProps>("bar", props, defaultProps);
```
becca-bailey, line 61: This is my initial idea for a pattern that will take the child's props and sync them with the provider props. For this basic use case, we are just setting the state of the provider using these props. But for a more advanced use case, like if there are multiple chart components nested inside a `VictoryChart` or `VictoryGroup`, we will need to add more logic in the provider for storing props from each child and getting derived values from each child's props. 

For example, if one child has data that ranges from 1 to 5 and another child has data that ranges from 3 to 10, the parent will need to know that the full domain of the data goes from 1 to 10. 

Also, if there is an edge case where there is a domain prop passed into both `VictoryBar` and `VictoryChart`, we will need to make a determination about which prop will override the other, and perhaps provide some sort of user feedback (like a console message) telling the user which prop takes precedence.

packages/victory-bar/src/v37/victory-bar.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  // TODO: Do we need to get this from context?
+  const formattedData = useData();
+
+  const children = formattedData.map((datum: Datum, i: number) => {
+    const { x, y, x0, y0 } = getBarPosition(
+      { domain, scale, horizontal },
+      datum,
+    );
+    const dataProps = {
```
scottrippey, line 76: Could we add a type to this?  Maybe `BarProps`?

packages/victory-bar/src/v37/victory-bar.tsx
```diff
@@ -0,0 +1,97 @@
...
+  });
+  const scale = useScale();
+  const domain = useDomain();
+
+  // TODO: Do we need to get this from context?
+  const formattedData = useData();
+
+  const children = formattedData.map((datum: Datum, i: number) => {
+    const { x, y, x0, y0 } = getBarPosition(
+      { domain, scale, horizontal },
```
scottrippey, line 73: Super minor, but could we create this object outside the loop?

scottrippey (approved):
Looks good! Some tiny comments but LGTM.

PR #2319, opened 2022-06-21
TypeScript: migrate `victory` to TypeScript

scottrippey:
# What
Migrates Victory to TypeScript.  
Updates unit test to use snapshots.
No other changes.


packages/victory/src/index.ts
```diff
@@ -57,32 +57,45 @@ export { VictoryVoronoi, Voronoi } from "victory-voronoi";
 export { VictoryBrushLine } from "victory-brush-line";
 export {
   VictoryBrushContainer,
+  // @ts-expect-error Missing types for this export
   BrushHelpers,
+  // @ts-expect-error Missing types for this export
   brushContainerMixin
```
scottrippey, line 63: There are a lot of exports in these packages that are missing type defs.  
Ignoring these via `// @ts-expect-error` will have no issues for now.
As these packages are migrated to TS, these `// @ts-expect-error` comments will have to be cleaned up! 👍

becca-bailey (approved):

PR #2321, opened 2022-06-23
Use build config for babel

### LINKED ISSUE #2320: Stop copying test files to `es` and `lib`
As a part of #2301, we moved our tests to the directories they are testing. This required a few config updates in order to get everything compiling correctly.

However, we ran into a conflict between tests and typescript. In order to package our generated type files with Victory, we are using the `--copy-files` flag in our babel CLI commands. This takes all files that have not been transformed with babel and copies them to the `es` and `lib` packaged output directories. 

We have some ability to configure which files are copied by ignoring certain files in the babel build process and using the `--no-copy-ignored` flag. However, `.d.ts` files are also ignored by babel, so this doesn't work for our use case.

As a temporary measure, we are copying all files, including test files to the compiled output, and we have set up both jest and npm to ignore these copied test files. 

Eventually, we should make sure these test files are not copied at all. This might involve some more advanced babel configuration, using another build tool, or removing our dependency on babel altogether once the typescript conversion is complete.

ryan-roemer:
Fixes #2320

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -134,9 +134,9 @@ module.exports = {
       "lerna version --no-git-tag-version --no-push --loglevel silly",
     // TODO: organize build scripts once build perf is sorted out
     "babel-es":
-      "cross-env BABEL_ENV=es babel src --out-dir es --config-file ../../.babelrc.js --extensions .tsx,.ts,.jsx,.js --source-maps",
+      "cross-env BABEL_ENV=es babel src --out-dir es --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js --source-maps",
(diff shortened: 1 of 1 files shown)


scottrippey (approved):
Looks good, thanks!

becca-bailey:
If this is working correctly, we should be able to remove some of the test ignore patterns in the jest config. I can push up another commit, or just do this after this change gets merged.

becca-bailey (approved):

ryan-roemer:
@becca-bailey your choice! I'll let you decide -- you can take this one over and merge as-is if you want to or add the test ignore changes then merge!

PR #2325, opened 2022-06-23
Remove dependency on use-context-selector

### LINKED ISSUE #2322: Module not found: Error: Can't resolve 'use-context-selector'
# Bug Reports

### Checklist

- [x] I have read through the [FAQ](https://formidable.com/open-source/victory/docs/faq) and [Guides](https://formidable.com/open-source/victory/guides/)

### The Problem

victory-core/es/victory-state seems recently to have added a dependence on `use-context-selector` without expressing this dependence in its package.json. As a result, webpack builds now fail.

Further complicating matters... the latest several versions of `use-context-selector` depend on `react >= 18`... so I'm not even sure how you all would express a dependence on this node module in a way that allowed for that kind of flexibility. oof. 

of note, patternfly, which uses you, is not yet react 18, and yet they express a `^` floating dependence on you that brings in this breaking change!

Full error message:
> Module not found: Error: Can't resolve 'use-context-selector' in '.../node_modules/victory-core/es/victory-state'

The fix, for react 17 users, is `npm install use-context-selector@1.1`; >1.1 have a react >= 18 dependence

### Reproduction

todo

becca-bailey:
This removes our dependency on use-context-selector in favor of vanilla context. Once we have a more fully-featured re-write, we can do some performance comparisons and measure whether something like `use-context-selector` provides noticeable performance gains, and reconsider introducing something similar. 

Fixes #2322

### DIFF
--- a/packages/victory-core/src/victory-state/victory-provider.tsx
+++ b/packages/victory-core/src/victory-state/victory-provider.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { createContext, useContextSelector } from "use-context-selector";
+import { createContext, useContext } from "react";
 import { D3ScaleFn, DomainTuple, ForAxes } from "../types/prop-types";
 import { FormattedDatum, getData } from "./helpers/get-data";
 import { getDomain } from "./helpers/get-domain";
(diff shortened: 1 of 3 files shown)


gksander (approved):
ezpz

packages/victory-core/src/victory-state/victory-provider.tsx
```diff
@@ -81,17 +81,12 @@ export function VictoryProvider({
...
-        "useVictoryContext must be used within a VictoryProvider",
-      );
-    }
-    return selector(context);
-  });
+  const context = useContext(VictoryContext);
+  if (!context) {
+    throw new Error("useVictoryContext must be used within a VictoryProvider");
+  }
+  return selector(context);
```
scottrippey, line 89: Well that was surprisingly easy :)

scottrippey (approved):

packages/victory-core/src/victory-state/victory-provider.tsx
becca-bailey: Abstraction for the win!

PR #2332, opened 2022-06-27
Local VSC settings to ignore built files in search

gksander:
`.vscode` is git-ignored, and I'm generally a fan of keeping IDE settings out of repos. However, I'm proposing we add a small amount of workspace/local settings to make search more useful. With the es/cjs build folders, searching for a file ends up showing triple the results. This PR excludes the es/lib results from that search.

image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "search.exclude": {
+    "**/node_modules": true,
+    "**/es": true,
+    "**/lib": true
+  }
+}


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "search.exclude": {
+    "**/node_modules": true,
+    "**/es": true,
+    "**/lib": true
+  }
+}
```
scottrippey, line 7: Since `.vscode` is ignored, would it make sense to add `!.vscode/settings.json` to the `.gitignore`, so that changes to this file will get picked up?

gksander: I generally prefer to leave it gitignore and then if you _need_ to change it, do a `git add .vscode/settings.json -f` (with the `-f` force flag) to explicitly add it. Otherwise, it's too easy for contributors' personal, local settings to get picked up in the git workflow (which I think we want to avoid).

PR #2333, opened 2022-06-27
Export all types from Victory

### LINKED ISSUE #2330: No longer export types in v36.5.2
**Is your feature request related to a problem? Please describe.**
I have used exported types such as VicoryTooltipProps or FlyoutProps to extend my own components.
However, they are no longer exported in v36.5.2.
Is there a plan to export them again in future version?

**Describe the solution you'd like**
Export `xxxProps`

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

scottrippey:
In #2319, I forgot to export types from other packages.  
This PR ensures we export all types alongside the matching components.

Fixes #2330


packages/victory/src/index.ts
```diff
@@ -1,11 +1,17 @@
 export {
   addEvents,
+  Arc,
+  type ArcProps,
```
scottrippey, line 4: FWIW, I had to use the `type` annotation here, because we have `--isolatedModules` enabled.  Kinda muddies up the water, but it's also kinda nice and explicit.

gksander (approved):
Nice!

Issue #2329, opened 2022-06-24
TypeError: Cannot read properties of undefined (reading 'animate')


jailsonpaca:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `victory-core@36.5.2` for the project I'm working on.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/victory-core/es/victory-util/hooks/use-animation-state.js b/node_modules/victory-core/es/victory-util/hooks/use-animation-state.js
index 8c3e1ef..063aa8f 100644
--- a/node_modules/victory-core/es/victory-util/hooks/use-animation-state.js
+++ b/node_modules/victory-core/es/victory-util/hooks/use-animation-state.js
@@ -43,7 +43,7 @@ export var useAnimationState = function () {
   }, [_setState]); // This is a copy of Wrapper.getAnimationProps
 
   var getAnimationProps = React.useCallback(function (props, child, index) {
-    if (!props.animate) {
+    if (!props || !props.animate) {
       return child.props.animate;
     }
 
@@ -76,7 +76,7 @@ export var useAnimationState = function () {
   }, [state, setState]); // This is a copy of Wrapper.setAnimationState
 
   var setAnimationState = React.useCallback(function (props, nextProps) {
-    if (!props.animate) {
+    if (!props || !props.animate) {
       return;
     }
 
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

wadehammes:
I am also getting this error now

enricoros:
Same error - thanks for this workaround. Please merge 🙏

gksander:
@jailsonpaca Can you provide a small example of where this error is thrown?

wadehammes:
> @jailsonpaca Can you provide a small example of where this error is thrown? I'd like to get this fix integrated and released, but want to see where it's being surfaced from and if there's anything else that needs to be fixed along with your proposed patch.

He has an example above of the two places where this is happening in the file `hooks/use-animation-state.js`

gksander:
Apologies, poor wording on my part. I meant a minimal reproducible example so it's easier for me to tell where in actual usage this error will occur.

gksander:
Thank you, @wadehammes! I'm seeing that

```jsx
<VictoryChart animate={{ duration: 2000, easing: "bounce" }}>
  <VictoryLine />
[...]

PR #2334, opened 2022-06-27
Fix for #2329 (use-animation-state crashing charts)

gksander:
I'm not exactly sure why this is _just now_ becoming an issue, but somehow a `props` value is `undefined` (I think on just first render) and it was throwing because of a property-lookup on `undefined`. Seems like _maybe_ because `usePreviousProps` is always going to return `undefined` on first render until the `useEffect` therein kicks in.

Nevertheless, this fix seems to resolve the issue reported in #2329.

### DIFF
--- a/packages/victory-core/src/victory-util/hooks/use-animation-state.js
+++ b/packages/victory-core/src/victory-util/hooks/use-animation-state.js
@@ -60,7 +60,7 @@ export const useAnimationState = (initialState = INITIAL_STATE) => {
   // This is a copy of Wrapper.setAnimationState
   const setAnimationState = React.useCallback(
     (props, nextProps) => {
-      if (!props?.animate) {
+      if (!props || !props.animate) {
         return;
       }
       if (props.animate.parentState) {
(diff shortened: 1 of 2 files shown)


packages/victory-core/src/victory-util/hooks/use-animation-state.js
```diff
@@ -60,7 +60,7 @@ export const useAnimationState = (initialState = INITIAL_STATE) => {
   // This is a copy of Wrapper.setAnimationState
   const setAnimationState = React.useCallback(
     (props, nextProps) => {
-      if (!props.animate) {
+      if (!props || !props.animate) {
```
scottrippey, line 63: Can we use optional chaining here too?

scottrippey (approved):
Thanks for creating the unit test!  I'm investigating too, to see how this might have broken.

packages/victory-core/src/victory-util/hooks/use-animation-state.js
wadehammes: You could also just destructure the props you need: `const {animate, children} = props;` and then check for those.

PR #2335, opened 2022-06-27
chore(typescript): improved type-safety of `use-animation-state`

scottrippey:
This TS conversion helped me find the root cause of #2329.
This PR does nothing more than improve type-safety.

### DIFF
--- a/packages/victory-chart/src/victory-chart.tsx
+++ b/packages/victory-chart/src/victory-chart.tsx
@@ -78,7 +78,7 @@ const VictoryChartImpl: React.FC<VictoryChartProps> = (initialProps) => {
 
     const mappedChildren = children.map((child, index) => {
       const childProps = assign(
-        { animate: getAnimationProps(props, child, index, "victory chart") },
+        { animate: getAnimationProps(props, child, index) },
         child.props,
       );
       return React.cloneElement(child, childProps);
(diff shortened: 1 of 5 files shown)


packages/victory-chart/src/victory-chart.tsx
```diff
@@ -78,7 +78,7 @@ const VictoryChartImpl: React.FC<VictoryChartProps> = (initialProps) => {
 
     const mappedChildren = children.map((child, index) => {
       const childProps = assign(
-        { animate: getAnimationProps(props, child, index, "victory chart") },
+        { animate: getAnimationProps(props, child, index) },
```
scottrippey, line 81: (invalid 4th parameter)

packages/victory-core/src/victory-util/hooks/use-animation-state.ts
```diff
@@ -4,26 +4,41 @@ import { defaults, some } from "lodash";
...
 import * as Transitions from "../transitions";
 
-const INITIAL_STATE = {
+const INITIAL_STATE: AnimationState = {
   nodesShouldLoad: false,
   nodesDoneLoad: false,
   animating: true,
 };
 
+type AnyObject = Record<string, any>;
```
scottrippey, line 13: I wish I would've used a type like this more frequently in TS conversions ... basically this just means "any not null".
Thinking out-loud, is this what `{}` is for?

Issue #2236, opened 2022-05-12
Infra: Get all deps updated


ryan-roemer:
Probably good timing in line with #2205 and/or we can do some piecemeal first:

- [x] Update "easy" deps
- [x] Audit and remove unused deps (especially after #2205)
- [x] Update `eslint` deps + all changed files
- [ ] update d3-array (and publish new version! (patch or minor bump?)
- [ ] Remove `.npmrc` thing needed with `strict-peer-dependencies=false` and actually fix peer deps
- [ ] Try to remove `package.json:pnpm.overrides` with updated packages (eslint-related)
- [ ] Search on `TODO(dependencies)`

```sh
$ pnpm outdated
# ...
````

carbonrobot:
Will be handled in new tickets

PR #2345, opened 2022-06-30
Infra: Switch monorepo tooling to wireit and pnpm

ryan-roemer:
This PR is a big infrastructure overhaul to switch us from lerna + yarn to wireit + pnpm.

## Why?

Our existing setup of yarn + lerna has the following undesirable things:

1. Installs are slow on yarn
2. Yarn1 isn't updated anymore and simple package dependency updates don't work out of the box. It's a huge pain to update anything.
3. Lerna is getting old, and while the NX team has agreed to support it now, our build is (1) slow, (2) not cached at all, and (3) opaque as to what you need to run task-wise to support other tasks.

So, let's welcome PNPM + Wireit!

1. PNPM is a fast installer and well supported for normal things like package upgrades.
2. Wireit allows fine-grained subtask caching and dependency specifications to allow us to: (1) specify dependencies so when you run `pnpm run jest` all the other things that need to happen magically happen, (2) caches sub-tasks so things that don't have changed file inputs don't re-run, and (3) has full GH actions CI support!

To get a sense of how much faster our build has become take a look at the CI times for this branch in https://github.com/FormidableLabs/victory/actions?query=branch%3Ainfra%2Fpnpm-wireit++ -- when no input files change our CI times are about 1 minute. When some things change, a couple minutes. When all things (or our base scripts) change, we take the full hit of a comparable existing Victory CI time of like 15-16 mins.

For the average Victory developer, if you're working in just one package, you can just run the project-level `pnpm run check` and have like a full build and everything work reasonably fast without needing to know more or run subtasks!

## Check it out

```sh
$ npm install -g pnpm
$ pnpm install
# This will be slow!
$ pnpm run check

# ... but the second time will be super fast! And as you change things subsequent runs should be very fast!
$ pnpm run check
```

## Implementation notes

- High level:
    - **Dependency graph architecture**: All of our tasks now use wireit to identify and run/cache dependencies. So, if you want to run jest, you just run `pnpm run jest` and don't need to worry about "what else needs to be built?" before that.
    - **Parallelization**: To best use wireit, we've refactored our tasks to be more package-specific where possible. E.g., we run lint in incremental mode per-package meaning that both Wireit (at package level) and eslint (at file level within package) re-execute on the narrowest unit of "changed files" possible.

- Demos: The demos originally had imports from `victory*/src/index` (source) which wasn't efficient or consistent because the transitive deps on other victory packages was on built files. I refactored these to be normal `victory(-<NAME>)` imports using built files.

- `src`:
    - Since we did a "fresh" install with pnpm, there were some updates in lint packages that meant source or tests needed small tweaks to continue passing/building.
    - Refactor self-references within a package to not use package name.
    - Fixed missing package dependencies uncovered as pnpm-based installs will fail on missing dependencies that are flattened and "accidentally work" in yarn/npm.

- Configs:
    - Webpack `resolve.alias`: Since we no longer hoist victory packages to root, we now add aliases for our webpack configs and storybook (which uses webpack) programatically.
    - Babel, Jest configs: Normalized the naming and location of these across normal and native ones.

- Incremental caching:
    - eslint

## Remaining work in this PR

Big tasks:

- [ ] **Victory Native**: We need to check that everything still works for reals in RN and for our demos.
- [x] **Chromatic**: We use chromatic mainly through GH actions which don't need changes. But we also have localdev tasks to test things out. Should be straightforward.
- [x] **Docs**: This is still an independent install via yarn. Should work, but let's check.
- [x] **Watch support**: Something reasonable for most workflows out of the gate. Ticket the rest.
    - [x] `start*`
    - [x] `check`
    - [x] `build`: Might combine with check (?)
- [x] **Fix scripts**: We have to pass different "fix" flags to things. Let's aggregate to individual and meta-tasks.

All the rest:

- [x] Update all tasks, workflows, and guides in CONTRIBUTING.md
- [x] **Configure wireit package locks** Needed customization for PNPM and we should identify which subtasks should have a package log dep. Add `pnpm-lock.yaml` as `files` field for some tasks.
- [x] Search on
    - [x] `yarn`
    - [x] `lerna`
    - [x] `nps` (direct uses)
    - [x] `TODO(wireit)`
    - [x] `TODO(pnpm)`

## Remaining work (ticketed)

**Incremental caching**: Investigate opportunities for tool-level incremental caching. https://github.com/FormidableLabs/victory/issues/2375

- `build:dist:*`: See https://webpack.js.org/configuration/cache/
- `format`: Need to update to `v2.7`. https://prettier.io/blog/2022/06/14/2.7.0.html
- `types`
- `jest`

**Various things**

- **Publish support**: Publish with changesets. https://github.com/FormidableLabs/victory/issues/2373
- **Update dependencies!** This is a huge pain with yarn, so we've been avoiding. After PNPM lands and everything works, let's finally update everything! https://github.com/FormidableLabs/victory/issues/2236
- Add code coverage, split out in `check:ci` and upload to CodeCov. https://github.com/FormidableLabs/victory/issues/2348
- Revisit how we're doing build watches for `start` and `start:ts`. https://github.com/FormidableLabs/victory/issues/2374
- Review legacy sub-projects that still use yarn: https://github.com/FormidableLabs/victory/issues/2376


scripts/sync-pkgs-wireit.js
```diff
@@ -0,0 +1,108 @@
...
+      addDeps("build:lib:cjs", `../${dep}:build:lib:cjs`);
+
+      // Webpack depends on ESM output from other packages.
+      addDeps("build:dist:dev", `../${dep}:build:lib:esm`);
+      addDeps("build:dist:min", `../${dep}:build:lib:esm`);
+    });
+
+    // Hack: move sideEffects back to end
+    delete pkg.sideEffects;
+    pkg.sideEffects = false;
```
scottrippey, line 114: I was thinking, if our `scripts` and `wireit` config are now generated, they should all be at the end.  So could we keep `sideEffects` "before the fold" if possible?

scottrippey (commented):
Looking good so far!  So excited about the initial results.
Excellent idea to automate the `wireit` config!

scripts/sync-pkgs-wireit.js
ryan-roemer: Yeah! I should remove them and move them all way higher up (which is more typical)

scripts/sync-pkgs-wireit.js
```diff
@@ -0,0 +1,108 @@
...
+  // Use the core package as the template for the rest.
+  const corePkg = require(`${PKGS_ROOT}/victory-core/package.json`);
+
+  for (let workspace of workspaces) {
+    const pkgPath = `${PKGS_ROOT}/${workspace}/package.json`;
+    const pkg = JSON.parse(await fs.readFile(pkgPath));
+
+    // Overwrite scripts and wireit configuration.
+    const scripts = JSON.parse(JSON.stringify(corePkg.scripts));
+    pkg.scripts = scripts;
```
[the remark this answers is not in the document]
ryan-roemer: Great idea

packages/victory-area/package.json
```diff
@@ -27,8 +28,138 @@
...
+      "files": [
+        "src/**",
+        "!src/**/*.test.*",
+        "../../.babelrc.build.js"
+      ],
+      "output": [
+        "es/**"
+      ],
+      "dependencies": [
+        "../victory-core:build:lib:esm",
```
scottrippey, line 70: In testing this out, I noticed that any changes to `victory-core` will result in a full-rebuild of all other packages.  This is not so great.

So let's talk about these dependency chains.
Technically, for `victory-area` to build, we don't actually need `victory-core` built.  So if we removed this dependency, that would fix the "full rebuild" problem.

However, obviously `victory-core` is a dependency that needs to be built for this package to work. So, where's the RIGHT spot to put this dependency?

It seems like `victory-core` is a "runtime" dependency.  It's needed for `jest`, and maybe it's needed for `build:dist`?

scottrippey: (and TypeScript will probably have its own full dependency chain on other types, but not necessarily on `build:esm` etc)

package.json
```diff
@@ -123,7 +103,248 @@
...
+  "scripts": {
+    "TODO_REFACTOR_start": "nps start",
+    "TODO_REFACTOR_start-ts": "nps start.ts",
+    "TODO_REFACTOR_storybook": "nps storybook.server",
+    "TODO_REFACTOR_build-storybook": "build-storybook",
+    "TODO_REFACTOR_chromatic": "chromatic",
+    "TODO_REFACTOR_chromatic-ci": "build-storybook && chromatic --ci --exit-zero-on-changes",
+    "check": "wireit",
+    "check:ci": "wireit",
+    "clean": "rimraf coverage && pnpm --recursive exec rimraf dist es lib",
```
ryan-roemer, line 125: I think this one might be better done as `rimraf coverage "packages/*/{dist,es,lib}`.

package.json
```diff
@@ -123,7 +103,248 @@
...
+    "clean:cache": "rimraf .wireit .eslintcache && pnpm run clean .wireit .eslintcache",
```
ryan-roemer, line 126: Same here.

package.json
scottrippey: Yep, that sounds better :)

package.json
scottrippey: Done

scottrippey:
I've been working hard on the TS support; see #2349

scottrippey:
I merged my TS branch here!  Let's continue working on things here (or in new feature-branches, if needed)

packages/victory-voronoi/package.json
```diff
@@ -28,7 +29,170 @@
...
+    "types:check": "wireit",
+    "types:create": "wireit",
+    "format": "wireit",
+    "lint": "wireit",
+    "jest": "wireit"
   },
-  "sideEffects": false
+  "wireit": {
+    "build": {
+      "dependencies": [
```
scottrippey, line 47: So here's my mental-model for our "dependency graph":
```mermaid
flowchart
    subgraph Key
    TLS{{entry scripts}} --> DS[Dependent scripts]
    end

    lint{{lint}}
    types:check{{types:check}}
    storybook{{storybook}}
    jest{{jest}}
    types:create{{types:create}}
    build{{build}}

    lint --> deps:types:create
    types:check --> deps:types:create
    types:create --> deps:types:create

    build --> types:create
    build --> build:lib:esm
    build --> build:dist
    build --> build:lib:cjs

    build:dist --> build:lib:cjs


    storybook -->  build:lib:cjs
    jest --> build:lib:cjs
```

scripts/sync-pkgs-wireit.js
```diff
@@ -0,0 +1,186 @@
...
+    const pkgPath = `${PKGS_ROOT}/${workspace}/package.json`;
+    const originalPkg = await readPkg(pkgPath);
+    const pkg = clone(originalPkg);
+
+    // Overwrite scripts and wireit configuration
+    pkg.scripts = clone(corePkg.scripts);
+    pkg.wireit = clone(corePkg.wireit);
+
+    // Clear out existing deps from victory-core
+    // TODO(wireit): Abstract and refactor this whole section better.
```
scottrippey, line 101: I agree with this comment, and have an idea...
Currently, we're pulling in our "source config" from `victory-core/package.json`, and then augmenting it here.

Instead, let's actually have the whole "source config" here in this file.  Here's a sample of what I'm thinking:

```js
function generateWireitConfig(pkg, rootPkg) {
  const isVictoryPackage = (p) => p.startsWith("victory") && p !== pkg.name; 
  const deps = Object.keys(pkg.dependencies).filter(isVictoryPackage);
  const devDeps = Object.keys(pkg.devDependencies).filter(isVictoryPackage);
  const rootDeps = Object.keys(rootPkg.devDependencies).filter(isVictoryPackage);
  return {
    "scripts": {
      "build": "wireit",
      "build:lib:esm": "wireit",
      // ...
      "types:create": "wireit",
      "format": "wireit",
      "lint": "wireit",
      "jest": "wireit"
    },
    "wireit": {
      "build": {
        "dependencies": [
          "build:lib",
          "build:dist",
          "types:create"
        ]
      },
      "build:lib:esm": {
        "command": "nps build:lib:esm",
        "files": [
          "src/**",
          "!src/**/*.test.*",
          "../../.babelrc.build.js"
        ],
        "output": [
          "es/**/*.js",
          "es/**/*.js.map"
        ],
        "dependencies": [
          ...deps.map(dep => `../${dep}:build:lib:esm`),
        ]
      },
      // ...
      "types:check": {
        "command": "nps types:pkg:check",
        "files": [
          "src/**/*.{ts,tsx}",
          "../../tsconfig.base.json",
          "tsconfig.json"
        ],
        "dependencies": [
          "types:create",
          ...[...deps, ...devDeps, ...rootDeps].map(dep => `../${dep}:types:create`),
        ],
        "output": []
      },
      "types:create": {
        "command": "nps types:pkg:create",
        "files": [
          "src/**",
          "!src/**/*.test.*",
          "../../tsconfig.base.json",
          "tsconfig.build.json"
        ],
        "output": [
          "es/**/*.d.ts",
          "es/**/*.d.ts.map",
          "lib/**/*.d.ts",
          "lib/**/*.d.ts.map"
        ],
        "dependencies": [
          ...deps.map(dep => `../${dep}:types:create`),
        ]
      },
      "jest": {
        "command": "nps jest:pkg",
        "files": [
          "src/**/*.test.*",
          "../../.babelrc.js",
          "../../jest-config.js",
          "../../test/jest-setup.js"
        ],
        "output": [],
        "dependencies": [
          "build:lib:cjs",
          ...[...deps, ...devDeps, ...rootDeps].map(dep => `../${dep}:build:lib:cjs`),
        ]
      }
    }
  }
}


```

scottrippey: (Merged)

.storybook/main.js
```diff
@@ -18,5 +33,6 @@ module.exports = {
     },
   ],
   // Use glob to locate the stories, because it ignores our circular dependencies:
+  // Remove .. ?
   stories: glob.sync("../**/*.stories.@(js|jsx|ts|tsx)", { cwd: __dirname }),
```
scottrippey, line 37: FWIW, I do think we'll still need this.
In my testing, I put an invalid `stories/INVALID.stories.ts` file.  The OOM error still happened, and it never even complained about the invalid syntax in the file.  So this tells me that it was stalling when "finding all stories" recursively, before it even checked what `imports` they had.

package.json
```diff
@@ -120,10 +103,398 @@
...
+    "lint:root": {
+      "command": "nps \"lint:base *.js .storybook scripts config demo stories test\"",
+      "files": [
+        ".eslintrc.js",
+        "*.js",
+        "scripts",
+        "config",
+        "demo",
+        "stories",
+        "typescript"
```
scottrippey, line 271: Should this list should be in-sync with the `command` list? I think we need `.storybook` and `test` in here, and I don't think `typescript` is a valid file/folder.

package.json
```diff
@@ -120,10 +103,398 @@
...
+        "./packages/victory-voronoi-container:build:lib:esm",
+        "./packages/victory-zoom-container:build:lib:esm"
+      ]
+    },
+    "storybook:build": {
+      "command": "build-storybook",
+      "files": [
+        ".storybook",
+        "stories"
+      ],
```
scottrippey, line 411: I think we also need `packages/**/*.stories.*` here

ryan-roemer: Oh! Thanks! I'll add that.

.storybook/main.js
ryan-roemer: Sounds good -- I've removed the `// Remove` comment

package.json
ryan-roemer: Um, ... YES!!!

For `test` I think VS Code autocompleted me to `typescript`. 😂 Fixing.

packages/victory-voronoi/package.json
ryan-roemer: Pretty close!!!

Storybook is webpack based, which means dependent on `build:lib:esm`. Similarly `build:dist` is webpack and also `:esm`

Jest is the only one using `build:lib:cjs`.

scripts/sync-pkgs-wireit.js
```diff
@@ -64,10 +64,10 @@ const updateRootPkg = async ({ allPkgs }) => {
   rootPkg.wireit = rootPkg.wireit || {};
   [
     { rootTask: "build", pkgTask: "build" },
+    { rootTask: "build:lib:esm", pkgTask: "build:lib:esm" },
```
scottrippey, line 67: I like how the script names match; maybe we should do this for the rest of the root scripts?  (remove the `:pkgs` part)

ryan-roemer: The issue is for some of our tasks like lint and others, we also have to add a root task for things outside of the packages. Build has no root tasks

packages/victory-area/package.json
ryan-roemer: I think this is resolvable now? 

For an import of `victory-area` to be correct in dev (tests, dev server, etc.), if a transitive victory dep has changes, we need the updated transpiled files. Fortunately, once the transitive dep (like `victory-core`) is rebuilt once, it's cached and shouldn't be a big deal elsewhere.

.eslintignore
```diff
@@ -9,3 +9,4 @@ artifacts
 tmp
```
scottrippey, line 9: Tried running `pnpm run check` ... unfortunately I got a JS OOM error.  Mostly-fresh everything, fresh `pnpm install` and ran both `clean:build` and `clean:cache`.
Then I did a "nuke"  (`git clean -xdf`) and now everything is passing.

So all this to say ... there's a hiccup somewhere, and we should keep our eyes out for it.

ryan-roemer: OK! Good not know! The only OOM I've gotten so far was the storybook thing that you successfully diagnosed and fixed.

package.json
```diff
@@ -120,10 +102,545 @@
...
+    "lint:pkgs": "wireit",
+    "lint:pkgs:fix": "wireit",
+    "lint:docs": "echo TODO FIGURE OUT INTEGRATION",
+    "jest": "wireit",
+    "jest:root": "wireit",
+    "jest:pkgs": "wireit",
+    "types:check": "wireit",
+    "types:create": "wireit",
+    "storybook:build": "wireit",
+    "storybook:server": "concurrently \"pnpm:build:lib:esm --watch\" \"start-storybook -p 6006\"",
```
scottrippey, line 151: This script could (eventually) use some improvement.  `concurrently` breaks Storybook's/webpack's logging, so every tiny change results in 500 log messages (`"[webpack.Progress]"`).  Just a minor nuisance.

ryan-roemer: Can you try again? I've added the `--raw` flag to all our concurrently use which should make this and the `start*` tasks work better in a terminal and not blast output.

scottrippey (approved):
Well I did a spot-check of several of the `--watch` workflows, and I'm thrilled!
It's awesome how we can just jump into `packages/victory-line` and run  `pnpm run jest` and it just works!  Pulls from all previous cache's.  Detects changes instantly, and reruns very fast.

LGTM!

package.json
scottrippey: Way better!  Messages are way more visible now, and errors don't get lost!

gksander (approved):
This is AWESOME. 🎈

PR #2347, opened 2022-07-01
victory-line T-T-T-TURBO!

scottrippey:
Implements `victory-line` using the new "Turbo" / v37 approach.


packages/victory-core/src/victory-util/with-default-props.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export type FCWithDefaultProps<TProps, TDefaultProps> = {
+  (props: TProps): React.ReactElement | null;
+  defaultProps: TDefaultProps;
+} & Pick<React.FC<TProps>, "propTypes" | "contextTypes" | "displayName">;
+
+/**
+ * Provides the maximum type-safety for the `Component.defaultProps` property.
+ */
+export function withDefaultProps<
```
scottrippey, line 15: Ignore this whole file.  It was an experiment to improve TS types.
It was successful, but, I later found a better way to do this, and included that solution in my "infrastructure" PR.

packages/victory-line/src/v37/victory-line.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const scale = useScale();
+  const domain = useDomain();
+  const dataProps = { ...props, data, scale, domain };
+
+  return (
+    <Clone element={props.groupComponent}>
+      <Clone element={props.dataComponent} {...dataProps} />
+    </Clone>
+  );
+};
```
scottrippey: Yeah ... soooo .... this was a little too easy.  That's a good thing, right?

packages/victory-core/src/victory-util/with-default-props.ts
scottrippey: Removed.

packages/victory-line/src/v37/victory-line.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const data = useData();
+  const scale = useScale();
+  const domain = useDomain();
+  const dataProps = { ...props, data, scale, domain };
+
+  return (
+    <Clone element={props.groupComponent}>
+      <Clone element={props.dataComponent} {...dataProps} />
+    </Clone>
+  );
```
scottrippey, line 101: What do you think about this `Clone` utility?  I prefer the JSX syntax, so that's why I chose this.  
It's a lot easier for me to understand than imperative code, especially the nesting.
Thoughts?

packages/victory-line/src/v37/clone.tsx
```diff
@@ -0,0 +1,13 @@
...
+  props: React.PropsWithChildren<
+    {
+      element: React.ReactElement<TProps>;
+      children?: React.ReactNode | React.ReactNode[];
+    } & TProps
+  >,
+) => {
+  const { children, element, ...rest } = props;
+  return React.cloneElement(element, rest as unknown as TProps, children);
+};
```
becca-bailey, line 13: Is this meant to help with the typings for cloneElement?

packages/victory-line/src/v37/victory-line.tsx
becca-bailey: I think this generally makes sense. I also found it hard to understand all the nesting with children in arrays, and this makes it more declarative how elements are nested.

becca-bailey: I think this also might be a place to handle some of the logic like `getSafeUserProps` in a centralized place.

becca-bailey:
This makes sense as a first step! We will still need to figure out how to do the two-way props syncing when there is more than one `VictoryLine` component, but that can be in the next PR.

packages/victory-line/src/v37/victory-line.tsx
```diff
@@ -0,0 +1,106 @@
...
+    | "groupComponent"
+    | "labelComponent"
+    | "samples"
+    | "sortKey"
+    | "sortOrder"
+    | "standalone"
+    | "theme"
+  >
+> = {
+  data: [
```
becca-bailey, line 69: I haven't tried running this yet, but does it work to pass default data in here? Or is the default data coming from the provider defaults?

scottrippey: Good question ...
In the next PR #2356 , I've done some work to make sure the `defaultProps` is assigned to the top-most wrapper.  This ensures that all "wrappers" (including the VictoryProvider, if needed) will be able to use these default props.  
And if there's a "parent provider" in the DOM, these default props will be properly sync'd with the parent.

packages/victory-line/src/v37/victory-line.tsx
scottrippey: Oh, that's a good idea.

gksander (approved):

PR #2349, opened 2022-07-05
Wireit: TypeScript

scottrippey:
# What
Adds TypeScript to build:
- `pnpm run types:create:pkgs` Creates all `.d.ts` files
- `pnpm run types:check:pkgs` Checks the TS for all packages, after building

# TODO
- [ ] Integrate into publish
- [ ] Determine how to handle "broken builds" (`wireit --continue`?)


ryan-roemer:
Great start!

Was able to run

```sh
$ pnpm -r --filter victory-core run types:create
```

and see the types output

Some directions for us from here:

- [ ] Split `types:create` to `types:create:cjs` and `types:create:esm`
- [ ] I really view this as part of "the build", so I think we're looking at making `types:create:cjs` a dependency of `build:lib:cjs` and `types:create:esm` a dependency of `build:lib:esm` in each workspace.
- [ ] Manually code no-ops in `victory-native` and `victory-vendor` (like is done for much of build and jest stuff already)
- [ ] I'm not sure if `types:create:esm` is a dependency of `build/div/{min,prod}`?
- [ ] Aggregate the `sync:*` tasks to maybe just `sync` and update the git dirty check in GH actions.

scripts/sync-pkgs-wireit.js
```diff
@@ -105,7 +116,7 @@ const updateLibPkgs = async ({ libPkgs }) => {
       }
     };
     const crossDeps = Object.keys(pkg.dependencies).filter((p) =>
-      p.startsWith("victory"),
```
ryan-roemer, line 108: Need to include the `victory` package as well as the `victory-*` packages.

scripts/sync-pkgs-wireit.js
```diff
@@ -130,10 +145,14 @@ const updateLibPkgs = async ({ libPkgs }) => {
       p.startsWith("victory"),
     );
     crossDevDeps.forEach((dep) => {
+      // Jest depends on CJS output
       addDeps("jest", dep, `../${dep}:build:lib:cjs`);
+
+      // TypeScript checking depends on types output from other packages.
```
ryan-roemer, line 151: I'd make this more specific to `the CJS types output`.

scripts/sync-pkgs-wireit.js
```diff
@@ -115,6 +126,10 @@ const updateLibPkgs = async ({ libPkgs }) => {
       // Webpack depends on ESM output from other packages.
       addDeps("build:dist:dev", dep, `../${dep}:build:lib:esm`);
       addDeps("build:dist:min", dep, `../${dep}:build:lib:esm`);
+
+      // TypeScript checking depends on types output from other packages.
+      addDeps("types:check", dep, `../${dep}:types:create`);
```
ryan-roemer, line 131: Need to split CJS, ESM

scripts/sync-pkgs-wireit.js
```diff
@@ -130,10 +145,14 @@ const updateLibPkgs = async ({ libPkgs }) => {
       p.startsWith("victory"),
     );
     crossDevDeps.forEach((dep) => {
+      // Jest depends on CJS output
       addDeps("jest", dep, `../${dep}:build:lib:cjs`);
+
+      // TypeScript checking depends on types output from other packages.
+      addDeps("types:check", dep, `../${dep}:types:create`);
```
ryan-roemer, line 152: Need to split CJS, ESM

scripts/sync-pkgs-wireit.js
scottrippey: So here's what I decide to do...
The esm and cjs builds are identical, so I consolidated to a single `types:create` script which will compile to the `es` folder, then copy to the `lib` folder. It's faster and way simpler.

ryan-roemer: I think that sounds fine. All the `build:lib:esm` and `build:lib:cjs` and `build:dist:*` tasks can just depend on `types:create`.

When you've pushed the change, we should both review the full `victory-core` task for that one `outputs`-wise, etc. to make sure that we don't clobber into the `:cjs` scripts...

packages/victory-core/package.json
```diff
@@ -122,6 +130,40 @@
...
+        "../victory-vendor:types:create",
+        "../victory-axis:types:create",
+        "../victory-bar:types:create",
+        "types:create",
+        "../victory-line:types:create"
+      ],
+      "output": []
+    },
+    "types:create": {
+      "command": "nps types:pkg:create",
```
ryan-roemer, line 150: Does tsc depend on lib/es ts files being present and built first?

packages/victory-core/package.json
```diff
@@ -47,7 +52,8 @@
     "build": {
       "dependencies": [
         "build:lib",
-        "build:dist"
+        "build:dist",
+        "types:create"
```
ryan-roemer, line 56: Our other tasks depend on things at the level of “build:lib:esm”, etc.

Do any of our other tasks need types created or is it solely for publishing?

packages/victory-core/package.json
scottrippey: Tsc does not care about the built es/cjs `*.js` files. It only cares about `.d.ts` files.
So the dependency chain only lists `types:create` tasks

packages/victory-core/package.json
scottrippey: Come to think of it, though, our `lint` task also depends on types, so I'll be sure to add it there too.

ryan-roemer: Sounds good

ryan-roemer:
For the "copy es to lib" approach, we're getting these output messages:

```
ℹ️ [packages/victory-voronoi-container:build:lib:cjs] Output files were modified since the previous run.
ℹ️ [packages/victory-group:build:lib:cjs] Output files were modified since the previous run.
ℹ️ [packages/victory-stack:build:lib:cjs] Output files were modified since the previous run.
ℹ️ [packages/victory-chart:build:lib:cjs] Output files were modified since the previous run.
ℹ️ [packages/victory-histogram:build:lib:cjs] Output files were modified since the previous run.
ℹ️ [packages/victory-canvas:build:lib:cjs] Output files were modified since the previous run.
```

Now things still run very quickly, but I do wonder if there's any higher level impact on things? Bouncing back and forth between `pnpm run build` and `pnpm run types:create:pkgs` has both repeating these warnings.

PR #2352, opened 2022-07-06
[WIP] fix(#2108) get label text size from dom rendering

### LINKED ISSUE #2108: Fix label spacing and text size approximation
This is an issue to overhaul how label size is determined. Right now, label size is approximated using the size of specific characters in specific fonts, but this does not adapt to custom fonts, non-Latin character sets, and RTL text. 

This solution might require some discussion - right now the only proposed fix for this would be to pre-render the text with a CSS rule like `visibility: hidden`, measure the text, and re-render the text with the correct spacing. Are there other solutions to this problem?

camray:
Fixes #2108

Uses DOM measurements for label width and height. This method proves to be more accurate and 27% more performant than the existing approximation solution as shown below, particularly for CJK and RTL languages.

Some technical limitations include:
- Lack of support for SSR & React Native
  - The `window` element does not exist for either so we continue to fall back to the approximation solution. 
  - It would be very helpful to get a React Native developer to add a similar solution that will work on mobile. 
- Testing is difficult and non-comprehensive
  - Since jsdom doesn't actually render and there is no window width/height, it cannot be measured in tests. 
  - I would love to hear solutions to this problem if any exist. Currently I'm testing that the element is added to the dom.

Screen Shot 2022-07-06 at 11 04 10 AM
Screen Shot 2022-07-06 at 11 03 08 AM
Screen Shot 2022-07-06 at 11 02 50 AM

### DIFF
--- a/packages/victory-core/src/victory-util/textsize.ts
+++ b/packages/victory-core/src/victory-util/textsize.ts
@@ -255,36 +255,35 @@ const _measureWithDOM = (
 
   const containerElement = document.createElementNS(
     "http://www.w3.org/2000/svg",
-    "g",
+    "text",
   );
   element.appendChild(containerElement);
 
   element.style.position = "fixed";
   element.style.top = "-9999px";
   element.style.left = "-9999px";
+  element.style.height = "auto";
+  element.style.width = "auto";
 
   document.body.appendChild(element);
 
   const lines = _splitToLines(text);
-  let aggregatedHeight = 0;
+  // let aggregatedHeight = 0;
   for (const [i, line] of lines.entries()) {
     const textElement = document.createElementNS(
       "http://www.w3.org/2000/svg",
-      "text",
+      "tspan",
     );
     const params = _prepareParams(style, i);
-    textElement.setAttribute("transform", `rotate(${params.angle})`);
-    textElement.setAttribute("fontSize", `${params.fontSize}px`);
-    textElement.setAttribute("line-height", params.lineHeight);
-    textElement.setAttribute("font-family", params.fontFamily);
-    textElement.setAttribute("letter-spacing", params.letterSpacing);
-    textElement.setAttribute("x", "0");
-    textElement.setAttribute("y", aggregatedHeight.toString());
+    textElement.style.fontFamily = "arial";
+    textElement.style.transform = `rotate(${params.angle})`;
+    textElement.style.fontSize = `${params.fontSize}px`;
+    textElement.style.lineHeight = params.lineHeight;
+    textElement.style.fontFamily = params.fontFamily;
+    textElement.style.letterSpacing = params.letterSpacing;
     textElement.textContent = line;
 
     containerElement.appendChild(textElement);
-
-    aggregatedHeight += textElement.getBBox().height;
   }
 
   const result = {
--- a/packages/victory-core/src/victory-util/textsize.ts
+++ b/packages/victory-core/src/victory-util/textsize.ts
@@ -186,7 +186,7 @@ export const convertLengthToPixels = (
   return result;
 };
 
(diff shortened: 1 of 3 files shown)


packages/victory-core/src/victory-util/textsize.ts
```diff
@@ -243,6 +246,63 @@ const _approximateTextHeightInternal = (text: string | string[], style) => {
   }, 0);
 };
 
+const _measureWithDOM = (
+  text: string | string[],
+  style?: TextSizeStyleInterface,
+): { width: number; height: number } => {
+  const element = document.createElement("div");
+  element.style.position = "absolute";
```
gksander, line 254: Thoughts on using `fixed` instead of `absolute` just to be _damn sure_ it won't be in the viewport?

gksander (approved):
This is sweet.

packages/victory-core/src/victory-util/textsize.ts
camray: Talked with Scott, might do the rendering in SVG instead to make double sure it's the correct size (+ RN support)

camray: Updated to fixed positioning

packages/victory-core/src/victory-util/textsize.ts
```diff
@@ -243,6 +246,72 @@ const _approximateTextHeightInternal = (text: string | string[], style) => {
...
+  document.body.appendChild(element);
+
+  const lines = _splitToLines(text);
+  for (const [i, line] of lines.entries()) {
+    const textElement = document.createElementNS(
+      "http://www.w3.org/2000/svg",
+      "tspan",
+    );
+    const params = _prepareParams(style, i);
+    textElement.style.fontFamily = "arial";
```
scottrippey, line 275: Can we remove this?

packages/victory-core/src/victory-util/textsize.ts
```diff
@@ -243,6 +246,72 @@ const _approximateTextHeightInternal = (text: string | string[], style) => {
...
+    textElement.style.letterSpacing = params.letterSpacing;
+    textElement.textContent = line;
+
+    containerElement.appendChild(textElement);
+  }
+
+  const { width, height } = containerElement.getBBox();
+  const result = { width, height };
+
+  element.remove();
```
scottrippey, line 289: I assume we create a lot of text elements; could we reuse the `element` between calls?  (obviously we'd have to remove the children here)

scottrippey (approved):
This looks amazing!  Great work!
I have some tiny suggestions, but nothing big.  Overall this looks great!

packages/victory-core/src/victory-util/textsize.ts
```diff
@@ -252,24 +321,24 @@ export interface TextSizeStyleInterface {
...
+    text: string | string[],
+    style?: TextSizeStyleInterface,
+    approximate = false,
+  ) => {
+    // Attempt to first measure the element in DOM. If there is no DOM, fallback
+    // to the less accurate approximation algorithm.
+    const isClient =
+      typeof window !== "undefined" &&
+      typeof window.document !== "undefined" &&
+      typeof window.document.createElement !== "undefined";
```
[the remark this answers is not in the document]
camray: Do you mean moving this out of this file and up to the label implementations, then pass in `approximate: bool`?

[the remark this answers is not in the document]
scottrippey: Actually I was just thinking moving the `const isClient...` to the root of this file.  This is super-micro, but super easy to change.

scottrippey:
Feel free to resolve my comments as you wish!

PR #2356, opened 2022-07-08
Turbo: Property Containers

scottrippey:
# What
A POC for "Property Containers", which eliminate some of the boilerplate for our components.


packages/victory-core/src/victory-state/victory-provider-maybe.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const updateChildProps = useVictoryContextMaybe(
+      (value) => value?.updateChildProps,
+    );
+    const hasParentProvider = !!updateChildProps;
+
+    React.useEffect(() => {
+      if (!hasParentProvider) return;
+
+      const id = Symbol("WithProvider");
+      updateChildProps(id, props);
```
becca-bailey, line 24: Is this id just a placeholder? I'm guessing we will need some way of providing unique IDs for each child.

packages/victory-core/src/victory-state/victory-provider.tsx
```diff
@@ -14,27 +14,42 @@ export interface ContextType {
...
+          result[id] = newProps;
+        }
+        return result;
       });
     },
     [],
   );
 
+  // TEMP: combine all props into a single result:
+  // TODO: instead, we should intelligently aggregate all these props
```
becca-bailey, line 45: Yep, this is essentially the approach I was thinking about. For some props like `data` we will need to return the specific data for each child, but for something like `domain` or `scale` we will need to get the total values for all the aggregated data.

packages/victory-core/src/victory-state/victory-provider.tsx
```diff
@@ -69,18 +84,39 @@ export function VictoryProvider({
...
+  );
+
+  return normalizedProps;
+}
+
+export function VictoryProvider({
+  children,
+  ...providerProps
+}: VictoryProviderProps) {
+  const value = useNormalizedProps();
```
becca-bailey, line 104: We will need some way of getting specific props for an individual component as well. Some will be the same, but some (like data or styles) might be specific to that component. So maybe we need to provide a `getProps` function through the provider value that accepts an ID?

becca-bailey (commented):
This is another step in the right direction! I like this approach for syncing and aggregating child props. The next step (I think) will be figuring out the logic for getting props for individual children in the case where there are multiple chart components.

packages/victory-core/src/victory-state/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./victory-provider";
 export * from "./types";
+export * from "./victory-provider-maybe";
```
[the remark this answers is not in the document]
scottrippey: No, it's kinda a weird name, but the idea is that sometimes a Victory component is inside a container, sometimes it's standalone.  This `VictoryProviderMaybe` HOC will detect if there's already a parent, and use it, but if it's missing, it'll add it.  So that's why I called it "maybe".

packages/victory-core/src/victory-state/victory-provider-maybe.tsx
scottrippey: FWIW, Grant and I have discussed some of the flaws with this `updateChildProps` approach -- esp. SSR support.
We have some ideas how to achieve a better flow, which will make this code obsolete in a future PR.

scottrippey:
The CI failures are due to newly-added exports, but this will be fixed by #2431

gksander (commented):
I like it. Looks like a CI failure tho, and maybe conflicts with [this PR](https://github.com/FormidableLabs/victory/pull/2431).

gksander (approved):
LGTM

PR #2382, opened 2022-07-25
Victory RN Demo: local package references

gksander:
I missed this in the wireit/pnpm migration, but the RN demo should actually use linked references to all packages (so that the core packages aren't installed from npm, but rather use the local versions). End result is a beautiful workflow like the one shown below.

victory-rn-demo


ryan-roemer (approved):
LGTM! If we decide to bring this into the pnpm workspaces in the future we can probably switch away from links if we want. But looks just fine this way for our use!

PR #2387, opened 2022-07-26
Ts migrate/victory box plot

ramenhog:
This PR migrates victory-box-plot over to TypeScript


packages/victory-box-plot/src/victory-box-plot.test.tsx
```diff
@@ -159,7 +159,7 @@ describe("components/victory-box-plot", () => {
           q3Component={
             <Border
               ariaLabel={({ datum }) => buildLabel(datum.x, datum._q3)}
-              tabIndex={({ index }) => index + 1}
+              tabIndex={({ index }) => (index as number) + 1}
```
gksander, line 162: I'm always hesitant to cast with `as`, since it's basically like saying "I _know_ that this is a number", without actually checking that it's a number.

In this case, `index` has a type of `string | number`, so I think it's more appropriate to do `+index + 1` to do a _runtime_ conversion to number instead of declaring that it's a number at the type level.

packages/victory-box-plot/src/helper-methods.tsx
```diff
@@ -152,7 +152,7 @@ export const getDomain = (props, axis) => {
   return Domain.createDomainFunction(getDomainFromData)(props, axis);
```
gksander, line 152: This file was converted to TS, but not modified a ton. So for an example like this:

```ts
const nanToNull = (val) => (isNaN(val) ? null : val);
```

that `nanToNull` function's argument `val` has a type of `any`, because we haven't declared a type for the arg. I'm wondering if we should be tightening down on this, but I'll defer to @scottrippey for what he thinks (not sure what we've been doing for previous TS conversions, and how strict we're being in internal/helper methods).

(Side note: the reason TS doesn't complain about this is because we have `"noImplicitAny": false,"` in our `tsconfig.base.json`. If we wanted to be a bit more strict in this regard, we'd flip that to `true` – but it's a pretty big commitment, especially in a codebase this large.)

gksander (commented):
This looks good! I have one suggestion, and a question regarding one of the files that was converted (and how strict we're being in regards to the helper functions) that I'm hoping Scott can weigh in on.

packages/victory-box-plot/src/helper-methods.tsx
ramenhog: @gksander while @scottrippey and I were pairing on some of the `victory-axis` conversions, he'd had mentioned that we can just port over the helper files for now to and prioritize getting the entire codebase in TS first and then strictly typing those functions in the future when we come across them in the future.

Happy to change things up if you think there are more concerning functions that should be typed now though!

gksander: That sounds fine to me, Scott's been leading the charge – so if it's good with him, it's good with me 😅

gksander (approved):

PR #2393, opened 2022-07-28
TS: migrate `victory-brush-container` and `victory-brush-line`

ramenhog:
This PR migrates `victory-brush-container` and `victory-brush-line` over to TypeScript


packages/victory-core/src/victory-util/selection.ts
```diff
@@ -83,7 +84,7 @@ export function getSVGEventCoordinates(
 export function getDomainCoordinates(
   props: Pick<VictoryCommonProps, "scale" | "horizontal">,
   domain?: DomainPropType,
-): DomainPropType {
+): { x: Tuple<number>; y: Tuple<number> } {
```
ramenhog, line 87: Not sure if this is the best way to resolve TS errors for `selectedCoords` and `fullCoords` [here](https://github.com/FormidableLabs/victory/pull/2393/files#diff-4db73b037461df1d7256e5f726db803033e0b4a673632a7e895d341df2b1e7a3R57)?

packages/victory-brush-container/src/brush-helpers.ts
```diff
@@ -1,8 +1,26 @@
...
+  getDefaultBrushArea(targetProps, cachedDomain, evt): any;
+  getSelectionMutation(point, box, brushDimension): any;
+  panBox(props, point): any;
+  constrainBox(box, fullDomainBox): any;
+  constrainPoint(point, fullDomainBox): any;
+  hasMoved(props): any;
+  onMouseDown(evt, targetProps): any;
+  onGlobalMouseMove(evt, targetProps): any;
+  onGlobalMouseUp(evt, targetProps): any;
+}
```
scottrippey, line 21: I don't think we need this interface ... the exported `const BrushHelpers` will automatically be able to infer this type from the implementation.

packages/victory-core/src/victory-util/selection.ts
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import { Collection, VictoryCommonProps } from "./index";
 import type { DomainPropType, ScaleXYPropType } from "../types/prop-types";
+import { Tuple } from "../../lib";
```
scottrippey, line 4: We probably shouldn't import from `lib` like this, probably a VS Code auto-mistake

ramenhog: Oops, dang you auto-import 😅

packages/victory-brush-container/src/brush-helpers.ts
```diff
@@ -452,7 +451,11 @@ const Helpers = {
         : targetProps.defaultBrushArea;
     const defaultBrushHasArea =
       defaultBrushArea !== undefined && defaultBrushArea !== "none";
-    const mutatedProps = { isPanning: false, isSelecting: false };
+    const mutatedProps: {
```
ramenhog, line 454: is it preferred to type inline like this or to create a separate type here? is there a general rule of thumb?

scottrippey: Normally, inline objects like this already have an inferred type, which is fine.  In this case, though, we needed to explicitly type because there's an extra prop that we might set later.  But the concept is the same -- the type has the same scope as the object, which is fine.

ramenhog: I see! got it, thanks 🙏

packages/victory-core/src/victory-util/selection.ts
scottrippey: I'd say, remove this type (and let it have an implicit return type).  Seems easier.

packages/victory-brush-container/src/brush-helpers.ts
```diff
@@ -150,7 +150,7 @@ const Helpers = {
       const fullDomainBox =
         targetProps.fullDomainBox || this.getDomainBox(targetProps, fullDomain);
       const constrainedBox = this.constrainBox(pannedBox, fullDomainBox);
-      return Selection.getBounds({ ...constrainedBox, scale, horizontal });
+      return Selection.getBounds({ ...constrainedBox, scale });
```
scottrippey, line 153: We should keep `horizontal` in here ... the `getBounds` method should be updated instead.

packages/victory-brush-container/src/victory-brush-container.tsx
```diff
@@ -1,11 +1,45 @@
...
+    props: VictoryBrushContainerProps,
+  ) => void;
+  onBrushDomainChangeEnd?: (
+    domain: { x: DomainTuple; y: DomainTuple },
+    props: VictoryBrushContainerProps,
+  ) => void;
+}
 
-export const brushContainerMixin = (base) =>
+type Constructor = new (...ars: any[]) => React.Component;
```
scottrippey, line 40:
```suggestion
type Constructor = new (...args: any[]) => React.Component;
```

scottrippey (approved):
Looks good! Some tiny comments, then LGTM

packages/victory-brush-container/src/brush-helpers.ts
ramenhog: Ahh, I didn't see `horizontal` was being used by `getDataCoordinates` initially. Updated!

PR #2394, opened 2022-07-29
TS migrate: `victory-scatter`

sblinde:
# Description
This PR migrates the `victory-scatter` package to TypeScript!


packages/victory-core/src/victory-util/add-events.tsx
```diff
@@ -62,6 +62,7 @@ export interface EventsMixinClass<TProps> {
   ): React.ReactElement;
   cacheValues<TThis>(this: TThis, obj: Partial<TThis>): void;
   getEventState: typeof Events.getEventState;
+  renderData(props: TProps);
```
sblinde, line 65: This line wasn't already added as part of previous PRs and was required to get this working.

packages/victory-scatter/src/victory-scatter.test.tsx
```diff
@@ -239,7 +239,7 @@ describe("components/victory-scatter", () => {
               ariaLabel={({ datum }) =>
                 `scatter point x: ${datum.x}, y:${datum.y}`
               }
-              tabIndex={({ index }) => index + 10}
+              tabIndex={({ index }) => (index as number) + 10}
```
ramenhog, line 242: sharing my own learning from a recent PR: since index here has a type of `string | number`, instead of explicitly casting to a number, it's safer to do a runtime conversion with `+index` here

gksander (approved):
Looks good!

One thing I wanted to point out is the use of `!`, the non-null assertion operator. This TS operator basically says, "I'm telling _you_, the compiler, that I know this value isn't `undefined` or `null`". It's basically a way of telling TS to shut up a bit. For test files, which is where you have it used here, this isn't really a big deal. In application code, I always advise _against_ this operator – as it waters down the usefulness of TS! It's generally better to do an actual check like `if (svg) { ....` than to do `svg!.foo...`.

No changes needed here, test files it's fine to tell TS to be quiet, just wanted to point this out as a nugget for when you go on to (hopefully) use TS in client projects!

packages/victory-scatter/src/victory-scatter.test.tsx
sblinde: Ah, thanks for this. Do you have an example of what you mean here?

sblinde:
> Looks good!
> 
> One thing I wanted to point out is the use of `!`, the non-null assertion operator. This TS operator basically says, "I'm telling _you_, the compiler, that I know this value isn't `undefined` or `null`". It's basically a way of telling TS to shut up a bit. For test files, which is where you have it used here, this isn't really a big deal. In application code, I always advise _against_ this operator – as it waters down the usefulness of TS! It's generally better to do an actual check like `if (svg) { ....` than to do `svg!.foo...`.
> 
> No changes needed here, test files it's fine to tell TS to be quiet, just wanted to point this out as a nugget for when you go on to (hopefully) use TS in client projects!

Thank you! Super helpful info!

packages/victory-scatter/src/victory-scatter.test.tsx
ramenhog: Yup! Here's what I ended up doing for a similar case: https://github.com/FormidableLabs/victory/pull/2387/files#diff-59245dd3946a6d7b3f4c439108b380ccb32954c2a205ecef8f02dcaa7369d0a0R162

sblinde: Oh! Haha that's super clear, awesome! Thank you so much, I'll do this!

PR #2396, opened 2022-07-29
TS Migrate: `victory-group`

sblinde:
# Description
This PR migrates `victory-group` to TypeScript.


packages/victory-group/src/victory-group.tsx
```diff
@@ -178,19 +199,22 @@ VictoryGroup.defaultProps = {
   theme: VictoryTheme.grayscale,
 };
 
-// We need to attatch the static properties to the memoized version, or else
+// TODO: This utility could be moved and used for other components
+const typedMemo: <T>(component: T, equalityCheck?: any) => T = React.memo;
```
sblinde, line 203: `Memo`izing doesn't play well with TypeScript, so this utility (thanks to others who suggested the idea) could help with getting some of the other components across the board, allowing a memoized component to be typed appropriately.

packages/victory-group/src/victory-group.tsx
```diff
@@ -178,19 +197,14 @@ VictoryGroup.defaultProps = {
...
-VictoryGroupMemo.displayName = "VictoryGroup";
-VictoryGroupMemo.role = "group";
-
-VictoryGroupMemo.expectedComponents = [
+const VictoryGroupMemo = React.memo(VictoryGroupBase, isEqual);
+export const VictoryGroup = VictoryGroupMemo as typeof VictoryGroupMemo &
+  VictoryComponentConfiguration<VictoryGroupProps>;
+VictoryGroup.displayName = "VictoryGroup";
+VictoryGroup.role = "group";
+VictoryGroup.expectedComponents = [
```
scottrippey, line 205: So I wanted to share what I did to fix the `React.memo` issue ...
In Victory, we do this weird thing where we assign "static" fields to our classes (eg. `role` and `expectedComponents`).
These have no special meaning to React, but we use them internally with some of our utilities.  

When we write classes, we can just do:

```
class VictoryAxis extends React.Component<VictoryAxisProps> {
  static role = "axis";
  static expectedComponents = [ "groupComponent" ];
  render() { ...render... }
}
```

If we convert this class component to a functional component, it's not too different:
```
const VictoryAxis = () => { ...render... }
VictoryAxis.role = "axis";
VictoryAxis.expectedComponents = [ "groupComponent" ];
```

This is the interesting TypeScript part.  When you declare a plain function, TypeScript will allow you to add "static fields" to that function, and won't give you any type errors.  It'll actually infer the type, based on these assignments!  So `VictoryAxis` will have an inferred type like:
```
interface VictoryAxis {
  (): React.Element;
  role: string;
  expectedComponents: string[];
}
```

But unfortunately, if we wrap our function in anything like `React.memo`, then it will no longer allow us to assign inferred fields.  It will only allow us to have the fields defined by `React.memo` (eg. `displayName, propTypes, defaultProps`).  So it makes it hard to assign our custom fields.

scottrippey: There are a couple of ways to fix this.

We can just use `Object.assign`, because it allows us to assign anything, like:
```
export const VictoryGroup = Object.assign(
  React.memo(VictoryGroupBase, isEqual),
  {
    role: "group",
    expectedComponents: [ "groupComponent" ],
  }
);
```

However, this has no type safety, so instead I'm casting `VictoryGroup` to include these extra fields.  I'm still looking for a cleaner way to handle this, since we do this same pattern in 2 other components (that have yet to be converted to TS).

gksander (approved):
LGTM

PR #2398, opened 2022-08-01
Bug: Add pnpm install --no-frozen-lockfile to version script

ryan-roemer:
In CI, `pnpm install` checks against lockfile for changes and fails. We _want_ lockfile changes in our version script, so explicitly allow.

See, e.g.: https://github.com/FormidableLabs/victory/runs/7618211238?check_suite_focus=true

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     }
   },
   "scripts": {
-    "version": "pnpm changeset version && pnpm install",
+    "version": "pnpm changeset version && pnpm install --no-frozen-lockfile",
     "publish": "nps git:tag && pnpm changeset publish --no-git-tag",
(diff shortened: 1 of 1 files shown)


ramenhog (approved):

gksander (approved):
👍

package.json
```diff
@@ -114,7 +114,7 @@
     }
   },
   "scripts": {
-    "version": "pnpm changeset version && pnpm install",
+    "version": "pnpm changeset version && pnpm install --no-frozen-lockfile",
```
scottrippey, line 117: I'm not quite sure why `pnpm install` is even necessary for a `version` script?

scottrippey (approved):

package.json
ryan-roemer: We need to commit lock file changes after version changes. Originally used —lock file-only but that had issues

scottrippey: Ah, that makes sense!

PR #2403, opened 2022-08-03
Chore: Update lint and corresponding files.

ryan-roemer:
Doing **only** lint fixes and impacted code changes to scope this as narrowly as possible. Part of #2236


packages/victory-bar/src/v37/victory-bar.stories.tsx
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable react/no-multi-comp */
 import * as React from "react";
 import VictoryBar from "./victory-bar";
-import VictoryChart from "victory-chart/lib/v37/victory-chart";
+import VictoryChart from "victory-chart/es/v37/victory-chart";
```
ryan-roemer, line 4: Storybook uses webpack which uses our `es/**` files not `lib/**`. Our wireit config thus depends on `build:lib:esm` so this is the proper refererence.

scottrippey (approved):
Everything looks good to me.  Ran locally without issues.

hnji (commented):
N

packages/victory-bar/src/v37/victory-bar.stories.tsx
hnji: - [ ] Yo

PR #2406, opened 2022-08-03
chore(build): Run checks even for doc-only changes

scottrippey:
Currently we skip CI checks for doc-only changes.
However, since these checks are "Required", it blocks doc-only PRs from merging.

This PR will run CI checks for any changes, even if doc-only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,13 +7,9 @@ on:
   push:
     branches:
       - main
-    paths:
-      - '**'
   pull_request:
     branches:
       - main
-    paths:
-      - '**'
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,8 @@
 name: CI
 
 # Runs build and test on:
-#   every push to main that has a change in a file
-#   every pull request with main branch as the base that has a change in a file
+#   every push to main
+#   every pull request with main branch as the base
 on:
   push:
     branches:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,21 +1,19 @@
 name: CI
 
 # Runs build and test on:
-#   every push to main that has a change in a file not in the docs folder
-#   every pull request with main branch as the base that has a change in a file not in the docs folder
+#   every push to main that has a change in a file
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,21 +1,19 @@
...
 # Runs build and test on:
-#   every push to main that has a change in a file not in the docs folder
-#   every pull request with main branch as the base that has a change in a file not in the docs folder
+#   every push to main that has a change in a file
+#   every pull request with main branch as the base that has a change in a file
 on:
   push:
     branches:
       - main
     paths:
```
ryan-roemer, line 10: Why don't we just remove the entire `paths` blocks?

scottrippey: Yeah that seems like a good idea!

ryan-roemer (approved):
LGTM!

scottrippey:
Side note:
Something is weird with the wireit cache.
On one of the runs, there was a no-cache-hit for the `16.x` build, and it took 12m (but the `14.x` and `18.x` builds had full-cache hits).
On the next run, there was a partial-cache-hit for the `18.x` build, and it took about 3m (but the `14.x` and `16.x` builds took only ~45s).

I'm wondering if we have a race condition due to our matrix? Does our wireit cache take into consideration the `${{ matrix.node-version }}`?  Something to look into, since this PR's checks should've been <1m.

.github/workflows/ci.yml
```diff
@@ -1,21 +1,15 @@
...
 
 # Runs build and test on:
-#   every push to main that has a change in a file not in the docs folder
-#   every pull request with main branch as the base that has a change in a file not in the docs folder
+#   every push to main
+#   every pull request with main branch as the base
 on:
   push:
     branches:
       - main
```
scottrippey, line 9: @gksander @ryan-roemer I just noticed that when merging to `main`, it runs this CI workflow as well as the "Release" workflow.
Should we eliminate this CI workflow on `main`, so that merge-to-main only runs the Release workflow?

gksander: In my experience, it's pretty common to run something like this even on push to `main`! I'm assuming to pick up any sort of "force push to main" scenarios, to make sure every commit to `main` definitely passes the expected tests.

PR #2410, opened 2022-08-05
Remove cyclic dependency

aashu331998:
Remove cyclic dependency in victory-core / victory-util of `index.ts` -> `selection.ts` -> `index.ts`

Screenshot 2022-08-05 at 2 58 43 PM


scottrippey (approved):

scottrippey:
Good find, thank you!

PR #2414, opened 2022-08-06
chore: Added "check" script for individual packages

scottrippey:
I love how our current "root" scripts are the same as our "package" scripts.  
This PR just adds the `check` script, so that I can do:
```
cd packages/victory-axis
pnpm run check
```
and also with the `--watch` option, which is a great workflow when working in a specific package.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,6 +103,24 @@ To develop against `victory-native`, please see the package [README](./packages/
 
 ### Tips and tricks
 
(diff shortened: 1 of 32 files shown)


ryan-roemer (approved):
Looks great! I've kicked the failing CI job to see if it passes on another run.

I don't need a re-review, but before merging would you mind:

- [x] Manually add in `check` to `victory-vendor` and `victory-native`
- [ ] Add a short tip in `CONTRIBUTING.md`

PR #2416, opened 2022-08-09
chore(typescript): improved `textAnchor` props for `victory-label`

### LINKED ISSUE #2361: Typings for VictoryLabel['textAnchor'] are incorrect
**Describe the bug**
As per [these typings ](https://github.com/FormidableLabs/victory/blob/main/packages/victory-core/src/victory-label/victory-label.tsx#L58
) the `textAnchor` can be set as a callback that accepts nothing
in the [docs](https://github.com/FormidableLabs/victory/blob/main/packages/victory-core/src/victory-label/victory-label.tsx#L58) and when actually debugging I can see that it actually accepts `CallbackArgs` as parameter

**Victory version**
36.4.1

scottrippey:
# What

This improves the callback signature for the `textAnchor` prop.
Fixes #2361

### DIFF
--- a/packages/victory-core/src/types/callbacks.ts
+++ b/packages/victory-core/src/types/callbacks.ts
@@ -51,3 +51,5 @@ export type PaddingOrCallback = number | BlockProps | VictoryPaddingCallback;
 export type OrientationOrCallback =
   | OrientationTypes
   | VictoryOrientationCallback;
+
+export type ValueOrCallback<TValue> = TValue | ((args: CallbackArgs) => TValue);
--- a/packages/victory-core/src/victory-label/victory-label.tsx
+++ b/packages/victory-core/src/victory-label/victory-label.tsx
@@ -17,6 +17,7 @@ import {
   NumberOrCallback,
   StringOrCallback,
   StringOrNumberOrCallback,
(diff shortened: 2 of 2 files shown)


packages/victory-core/src/types/callbacks.ts
```diff
@@ -51,3 +51,5 @@ export type PaddingOrCallback = number | BlockProps | VictoryPaddingCallback;
 export type OrientationOrCallback =
   | OrientationTypes
   | VictoryOrientationCallback;
+
+export type ValueOrCallback<TValue> = TValue | ((args: CallbackArgs) => TValue);
```
scottrippey, line 55: IMO, I find the `xxxOrCallback` types to be very confusing ... like there's `StringOrCallback` and `StringOrNumberOrCallback` and `NumberOrCallback`.

This new utility can replace all of these, like:
```diff
-    angle?: StringOrNumberOrCallback;
+    angle?: ValueOrCallback<string | number>;
```

If we decide to go this route, internally, then maybe we could mark the old ones `@deprecated`.  But I'll leave that for another PR.

gksander (approved):
Nice 👍

Issue #2402, opened 2022-08-03
Typescript returns an error due to `datum` not being in `CoordinatesPropType`


stur86:
**Describe the bug**
`CoordinatesPropType`, as passed to the method used in `cursorLabel`, contains a `datum` member; however it seems to be missing in the type definition, and this causes Typescript to experience compilation errors.

**Victory version**
NPM: 8.15.1
Victory: 36.5.3

**To Reproduce**
1. Create a Typescript project using Victory
2. Create a `VictoryChart` object containing a `VictoryCursorContainer` one
3. Set the `VictoryCursorContainer` property `cursorLabel` with a function like `({ datum }) => `${datum.x}`

**Expected behavior**
This should compile and work well, but it fails due to `datum` not being found as part of `CoordinatesPropType`.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 103

**Additional context**
Currently using a project initialised with `create-react-app`.

scottrippey:
Our type definition for this prop was incorrect, and has been fixed in #2419 👍 Thanks for finding this!

scottrippey:
Fixed in `36.6.1` 👍

PR #2419, opened 2022-08-10
TS: Migrate victory-cursor-container

scottrippey:
# What 

Migrate `victory-cursor-container` to TS.

This includes a fix for #2402.


gksander (approved):
🚀 

Perhaps we also throw a changeset in here?

Issue #2338, opened 2022-06-28
Unable to install an older version of Victory due to carets in internal versions


tuncaulubilge:
# Questions

### Checklist

- [x] I have read through the [FAQ](https://formidable.com/open-source/victory/docs/faq) and [Guides](https://formidable.com/open-source/victory/guides/)

- [x] I am using the latest version of Victory

- [x] I've searched open issues to make sure I'm not opening a duplicate issue

### Tech

Using Yarn v1 (1.22.18)

### The Problem

We need to use an older version of Victory Native but we are unable to do so due to the carets in internal package versions. 

Even if we pin the version of victory-native to a certain version (lets say `36.5.2`), we still end up with `v36.5.3` of all the other victory sub-packages, such as `victory`, `victory-core`, `victory-line` etc. This is due to the usages of carets in version definitions of `victory-native`'s package.json: https://github.com/FormidableLabs/victory/blob/main/packages/victory-native/package.json#L26-L52

### Expected Behaviour

We would expect the packages to use pinned versions of all the internal dependencies to other victory packages, so if we pin a version to a certain number, all the other sub-packages would use the same version as well. Currently, this is not possible.

### Why do we need to pin the versions?

We have noticed recently that a minor version bump (from 36.4 to 36.5 iirc), have introduced API changes in victory, which was breaking for us. Also, some recent changes on 36.5.3 does require us to have the newer version of Typescript, which is not a migration we want to take on right away.

### Can we not rely on yarn.lock or `resolutions` in yarn?

Short answer is no. We have a multi-repo architecture so this would require us to incur a huge tech debt across all the repos, adding 20+ resolutions to each repo for every victory package, which is not scalable in terms of maintenance.

gksander:
Hi @tuncaulubilge – do you recall what sort of (breaking) API changes you noticed from the `36.4 → 36.5` bump? Also, what specific version of `victory-native` is most appropriate for your use-case?

We're currently thinking about:

1. Should we be version-pinning in `victory-native` moving forward?
2. Even if we version-pin moving forward, that doesn't help you out much since you're limited to `<36.5.3` anyways. Also thinking about how we could produce a patch/tagged release to help in your situation.

tuncaulubilge:
[...]

PR #2424, opened 2022-08-12
Remove carets from internal versions

tuncaulubilge:
Removes flexible versions from all the packages to address issue: https://github.com/FormidableLabs/victory/issues/2338

Flexible versions with carets or tilde are causing issues when installing older versions of victory


gksander:
Hi @tuncaulubilge – thank you for putting this together! One concern with pinning _everything_ across all packages is that if an end-user is installing packages individually, and happens to install different patch versions of two separate packages – they won't be able to share `victory-core` packages as intended, due to the inflexibility of version pinning.

However, that's not really a concern for the main `victory` and `victory-native` packages, since those just expose all of the individual packages. What do you think about just pinning all of the package dependencies in `victory` and `victory-native` (and not the sub-packages, like `victory-area` etc)? Would that make sense for your use-case?

tuncaulubilge:
Hi @gksander , if we don't use pinned versions in sub-packages (such as `victory-area`), we would have resolution conflicts, and potentially end up with multiple versions of another sub-package.

Let's assume `victory` pins to version 35.0.0 of `victory-area` and `victory-bar`, but version 35.0.0 of `victory-area` depends on version `^35.0.0` of `victory-bar` (with the caret). We could end up with multiple versions of `victory-bar` installed, as `^35.0.0` would resolve `35.1.2`. Yarn v1 is not smart enough to resolve both `35.0.0` and `^35.0.0` to 35.0.0

As for your concern for end-users installing packages individually, we can address the problem by utilising `peerDependencies`. Each sub-package should define a peer dependency to the exact same version of the victory core package, so it'd throw an error during install if you are installing two separate versions. To implement this, we might need to update the `changesets` script to regularly change the version of the peer dependency.

PR #2431, opened 2022-08-18
Turbo: no exports for now

scottrippey:
# What

- Isolates all `v37` code in `victory-core`, putting it into a `v37` folder
- No longer exports any of the v37 code from the root 
- I'm assuming that this shouldn't have ever been "public", since it's not quite usable yet, so it seems like no major version bump is needed


scottrippey:
@becca-bailey @gksander Up for discussion: should we hide all `v37/turbo` features from production for now?
All code will still be compiled and published, but just not included in the root exports.
Eg.
```diff
- import { useDomain } from "victory-core";
+ import { useDomain } from "victory-core/es/v37";
```

gksander (approved):
Nice!

Issue #2346, opened 2022-07-01
Webpack warns about not finding source map files


danielberndt:
When updating to victory v36.5.1, webpack issues a lot of warnings (159 in my case) with these contents:

```
WARNING in ../node_modules/victory/es/index.js
Module Warning (from ../node_modules/react-scripts/node_modules/source-map-loader/dist/cjs.js):
Failed to parse source map from '[...]/node_modules/victory/es/index.js.map' file: Error: ENOENT: no such file or directory, open '[...]/node_modules/victory/es/index.js.map'
```

I'm using create react app 5.0.1 for my app. I.e. it's based on webpack 5.73.0

This was likely introduced with #2307

marlonschlosshauer:
I have the same issue on a fresh React (18.2) install with Victory v36.5.3. Downgrading didn't fix it either (tried v36.0). Also seems to brick React's hot-reloading.

siulca:
Same issue here on Victory v36.5.2

miriyas:
Same issue here on Victory v36.5.3

Chee7ah:
➕

gbspr:
Any update to this bug or workaround?

scottrippey:
I am currently investigating, and working on a fix.  Thank you for reporting!

scottrippey:
Fixed in `36.6.3`, thanks for reporting!

PR #2307, opened 2022-06-17
Chore: generate source maps

scottrippey:
This PR generates source maps, to improve the local development experience.  "Go to definition" now works as expected, even from linked packages.

These source maps are not published to NPM.

PR #2432, opened 2022-08-18
Chore(build): do not build JS sourcemaps

scottrippey:
# What
We introduced source-map generation to improve Victory development (#2307).  
However, we intentionally did not publish these maps to NPM due to excessive file size (several MB).

This PR disables source-maps from the `.js` builds.  This prevents the `//# sourceMappingURL=index.d.ts.map` comments from being generated and published.  

This PR does NOT disable `.d.ts.map` generation.  This allows the developer workflow to stay awesome, since all of our core packages have been converted to TS.  (these map files are still not published to NPM)

This will address #2346

### DIFF
--- a/package-scripts.js
+++ b/package-scripts.js
@@ -31,9 +31,9 @@ module.exports = {
     // Build.
     // - Libraries
     "build:lib:esm":
-      "cross-env BABEL_ENV=es babel src --out-dir es --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js --source-maps",
+      "cross-env BABEL_ENV=es babel src --out-dir es --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js",
     "build:lib:cjs":
-      "cross-env BABEL_ENV=commonjs babel src --out-dir lib --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js --source-maps",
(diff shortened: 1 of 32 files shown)


package-scripts.js
```diff
@@ -31,9 +31,9 @@ module.exports = {
     // Build.
     // - Libraries
     "build:lib:esm":
-      "cross-env BABEL_ENV=es babel src --out-dir es --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js --source-maps",
+      "cross-env BABEL_ENV=es babel src --out-dir es --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js",
     "build:lib:cjs":
-      "cross-env BABEL_ENV=commonjs babel src --out-dir lib --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js --source-maps",
+      "cross-env BABEL_ENV=commonjs babel src --out-dir lib --config-file ../../.babelrc.build.js --extensions .tsx,.ts,.jsx,.js",
```
scottrippey, line 36: One problem with this PR, potentially, is that I don’t think our `wireit` cache will be invalidated by this change.  I might need to add this file to the wireit inputs.

scottrippey:
I did a spot-check, and the only build difference is the lack of the `.map`  file and the lack of the `//# sourceMappingURL=index.d.ts.map` comments! 👍

ryan-roemer (approved):
Thanks!

Issue #2360, opened 2022-07-12
Data Accessors using a function gives Typescript compiling errors


jacovanwijk:
# Bug Reports

### Checklist

- [x] I have read through the [FAQ](https://formidable.com/open-source/victory/docs/faq) and [Guides](https://formidable.com/open-source/victory/guides/)

- [x] I am using the latest version of Victory

- [x] I've searched open issues to make sure I'm not opening a duplicate issue

### The Problem

Since version 3.5.* using the Data Accessors on either bar or line using a function gives Typescript compiling errors.
It think the following commit caused the issue: https://github.com/FormidableLabs/victory/commit/43baeadb734bcb6c03e9abb4806fbe625f49ea4e#diff-066d807c91e06b799cde9263f3800161e2382910e1f5e259fd8c8a35eab1ba63R14

Changing object to any would fix the error, although you probably want to avoid any.

### Reproduction

Sandbox
https://codesandbox.io/s/victory-brush-zoom-chart-forked-ucmswz?file=/src/components/Chart.tsx

scottrippey:
I think, at this point, using `any` is the best bet.  
We don't (yet) support generics on our components (eg. where we could infer the `data` type and use that for the `datum` parameter).
And we don't know anything about the shape of `datum`.

So it makes sense to type as `any`, which would also allow the consumer to explicitly specify their own type.

scottrippey:
Published as `36.6.4`! Thanks for reporting!

PR #2436, opened 2022-08-19
Fix (TypeScript): give `datum` an `any` type

scottrippey:
# What

The common props `x`, `y`, `y0`, and `sortKey` allow for "accessor" functions, like:
```
<VictoryLine
  data={data}
  x = {(datum) => datum.xValue}
  y = {(datum) => datum.yValue}
/>
```
Currently, we have the `datum` arg typed as `object`, which doesn't allow for any property access nor strong types (see #2360).

This PR allows for the type to be `any`.  This seems like the best approach, since we currently don't know the shape of Datum, nor do we support generics with any of our components.
This will allow the above code to function properly, as well as support manually-added types (like `(datum: MyData) => datum.xValue`).

### DIFF
--- a/packages/victory-core/src/types/callbacks.ts
+++ b/packages/victory-core/src/types/callbacks.ts
@@ -51,5 +51,3 @@ export type PaddingOrCallback = number | BlockProps | VictoryPaddingCallback;
 export type OrientationOrCallback =
   | OrientationTypes
   | VictoryOrientationCallback;
-
-export type ValueOrCallback<TValue> = TValue | ((args: CallbackArgs) => TValue);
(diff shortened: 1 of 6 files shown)


packages/victory-core/src/types/callbacks.ts
```diff
@@ -51,5 +51,3 @@ export type PaddingOrCallback = number | BlockProps | VictoryPaddingCallback;
 export type OrientationOrCallback =
   | OrientationTypes
   | VictoryOrientationCallback;
-
-export type ValueOrCallback<TValue> = TValue | ((args: CallbackArgs) => TValue);
```
gksander, line 55: Looks like `types:check` is failing because of a couple instances where this is still being used.

scottrippey: Oh weird, I don't know how I missed those.

gksander (approved):
LGTM

Issue #2439, opened 2022-08-22
Type Errors: cannot find module "immutable"


scottrippey:
Fix the following type errors, which still remain on the latest build:

```
../../common/temp/node_modules/.pnpm/victory-core@36.6.4_react@17.0.2/node_modules/victory-core/lib/victory-util/immutable.d.ts(1,42): error TS2307: Cannot find module 'immutable' or its corresponding type declarations.
../../common/temp/node_modules/.pnpm/victory-vendor@36.6.4/node_modules/victory-vendor/d3-scale.d.ts(5,15): error TS6137: Cannot import type declaration files. Consider importing 'd3-scale' instead of '@types/d3-scale'.
../../common/temp/node_modules/.pnpm/victory-vendor@36.6.4/node_modules/victory-vendor/d3-shape.d.ts(5,15): error TS6137: Cannot import type declaration files. Consider importing 'd3-shape' instead of '@types/d3-shape'.
../../common/temp/node_modules/.pnpm/victory-zoom-container@36.6.4_react@17.0.2/node_modules/victory-zoom-container/lib/index.d.ts(31,15): error TS7006: Parameter 'evt' implicitly has an 'any' type.
../../common/temp/node_modules/.pnpm/victory-zoom-container@36.6.4_react@17.0.2/node_modules/victory-zoom-container/lib/index.d.ts(32,18): error TS7006: Parameter 'evt' implicitly has an 'any' type.
../../common/temp/node_modules/.pnpm/victory-zoom-container@36.6.4_react@17.0.2/node_modules/victory-zoom-container/lib/index.d.ts(37,19): error TS7006: Parameter 'ctx' implicitly has an 'any' type.
../../common/temp/node_modules/.pnpm/victory-zoom-container@36.6.4_react@17.0.2/node_modules/victory-zoom-container/lib/index.d.ts(39,13): error TS7006: Parameter 'props' implicitly has an 'any' type.
../../common/temp/node_modules/.pnpm/victory-zoom-container@36.6.4_react@17.0.2/node_modules/victory-zoom-container/lib/index.d.ts(48,3): error TS2749: 'RawZoomHelpers' refers to a value, but is being used as a type here. Did you mean 'typeof RawZoomHelpers'?
```

_Originally posted by @kaiyoma in https://github.com/FormidableLabs/victory/issues/2358#issuecomment-1221176693_

scottrippey:
Fixed in `36.6.5`

PR #2440, opened 2022-08-22
TS: more type improvements

scottrippey:
# What

Victory is currently causing trouble with certain TS configurations.
This addresses the known issues in #2439.


gksander (approved):
LGTM

PR #2441, opened 2022-08-22
Turbo: data aggregator

scottrippey:
# What
This "data aggregator" approach is to replace the Context-based approach currently used by the Turbo refactor.

# The problem
Victory components can be nested inside a parent container:
```jsx
<VictoryChart>
  <VictoryLine data={data} />
  <VictoryBar data={data} />
  <VictoryScatter data={data} />
</VictoryChart>
```

The parent container, in this example, needs to aggregate all the data from the children, to determine the range, domain, scale, etc.

In the current Turbo refactor, we did this via Context; each child reports its data to the parent.  This happens in a `useLayoutEffect`, which suffers from major drawbacks -- primarily that it breaks SSR support.

# The "Data Aggregator" Approach
This approach is derived from Victory's pre-Turbo behavior.  When a parent is rendered, it will actually **traverse the `children`** and aggregate all data before rendering.
This works with SSR, and prevents unneeded rerendering.  

This also gives us a chance to normalize the data, and provide a simple API for accessing "normalized" data.


scottrippey:
@carbonrobot This PoC is complete. The approach worked great, but this only applies to the turbo refactor. I suggest we get this merged, as it vastly improves the turbo codebase. Then we can decide what to do with it along with the rest of turbo.

test/jest-config.js
```diff
@@ -7,6 +7,12 @@ module.exports = {
   preset: "ts-jest",
   testEnvironment: "jsdom",
   rootDir: ROOT,
+  globals: {
+    "ts-jest": {
+      tsconfig: path.resolve(ROOT, "./test/tsconfig.json"),
+      isolatedModules: true, // Disables type-checking
+    },
+  },
```
carbonrobot, line 15: Any reason we want to do this, or was this for POC only?

package.json
```diff
@@ -151,7 +151,8 @@
     "types:create": "wireit",
     "storybook:build": "wireit",
     "build-storybook": "pnpm run build:lib:esm && pnpm exec build-storybook",
-    "storybook:server": "concurrently --raw \"pnpm:build:lib:esm --watch\" \"start-storybook -p 6006\"",
+    "storybook:server": "pnpm build:lib:esm && concurrently --raw \"pnpm:build:lib:esm --watch\" \"start-storybook -p 6006\"",
```
carbonrobot, line 154: This command looks like it runs twice, is that intentional?

test/jest-config.js
scottrippey: This prevents type checking while running tests in watch mode. Since we run type checks in CI, and the IDE shows type errors, it's ok for us to disable this.

package.json
scottrippey: Yes, this is intentional. The first command warms up the cache, and the second command keeps it updated.

test/jest-config.js
carbonrobot: Sounds good, as long as there is a still a way to Type check before pushing to the repo. Otherwise we have to wait on a pretty long CI run

carbonrobot (commented):
Could you update and sync the pnpm merge conflict?

test/jest-config.js
scottrippey: Yes, we have `pnpm run types:check` which also runs as part of `pnpm run check`, in addition to IDE support, that should be sufficient.

Issue #2463, opened 2022-09-12
TypeScript error with VictoryLineProps interface


dlabrecq:
**Describe the bug**
Encountering a TypeScript error with the `animate` prop of `VictoryLineProps` .  The `VictoryLineProps` interface appears to have been changed to only accept a boolean only?

![Screen Shot 2022-09-12 at 1 27 04 PM](https://user-images.githubusercontent.com/17481322/189718168-8d6eb6bc-f255-4d32-bdaf-f6792600dc49.png)

**Victory version**
36.6.6

**Expected behavior**
Per the docs below, the prop should accept a boolean or an object.

https://formidable.com/open-source/victory/docs/victory-line#animate

**Screenshots**

![Screen Shot 2022-09-12 at 1 16 31 PM](https://user-images.githubusercontent.com/17481322/189716234-2d2a0a4f-674f-4ce0-8c6a-498ddda0861d.png)

 - OS: macOS Monterey
 - Version 12.4

dlabrecq:
Fixed via https://github.com/FormidableLabs/victory/pull/2464

scottrippey:
Fixed in `36.6.7`!

PR #2464, opened 2022-09-13
VictoryLine's animate type overrides VictoryCommonProps

dlabrecq:
This fixes a TypeScript error where `VictoryLine`'s `animate` type inadvertently overrides `VictoryCommonProps`. According to the  `VictoryLine` docs, the `animate` prop should accept `boolean | AnimatePropTypeInterface`.

Closes: https://github.com/FormidableLabs/victory/issues/2463

### DIFF
--- a/packages/victory-line/src/victory-line.tsx
+++ b/packages/victory-line/src/victory-line.tsx
@@ -150,5 +150,4 @@ export interface VictoryLineProps
   interpolation?: InterpolationPropType | Function;
   samples?: number;
   style?: VictoryStyleInterface;
-  animate?: boolean;
 }


packages/victory-line/src/victory-line.tsx
```diff
@@ -150,5 +150,4 @@ export interface VictoryLineProps
   interpolation?: InterpolationPropType | Function;
   samples?: number;
   style?: VictoryStyleInterface;
-  animate?: boolean;
```
scottrippey, line 153: Tested locally, works great!  Just a mistake from the TS migration, thanks for finding this.

scottrippey (approved):

PR #2466, opened 2022-09-14
Enable Chromatic again

scottrippey:
This reverts commit 8ac47935bf43009e9a3169108f22fa51c54b4ecb.

# TODO:
- [x] Ensure the "Victory Turbo" stories aren't being captured yet

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -7,7 +7,8 @@ on:
     branches:
       - main
     paths:
-      - "stories/**"
+      - 'stories/**'
+      - 'packages/**'
 
 jobs:
   chromatic:
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -8,11 +8,10 @@ const STORIES = path.resolve(ROOT, "stories");
 module.exports = {
   webpackFinal: async (config) => {
     // Read all the victory packages and alias.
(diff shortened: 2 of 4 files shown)


.github/workflows/chromatic.yml
```diff
@@ -7,7 +7,8 @@ on:
     branches:
       - main
     paths:
-      - "stories/**"
+      - 'stories/**'
+      - 'packages/**'
```
ryan-roemer, line 11: Should this be:

```suggestion
      - 'packages/*/src/**/*.stories.*'
```

for specificity?

ryan-roemer: 🤔 or should we just enable it as is for really any `src` changes in case that impacts Storybook?

scottrippey: This pattern is actually used for marking _when_ to trigger Chromatic, which we want to do any time we change `packages/**`.

We do have a glob pattern in `.storybook/main.js` which finds all stories to include. 
I'll update it to EXCLUDE all the `v37` stories that shouldn't be included yet.

scottrippey:
Ironically, since this PR doesn't change any `packages/**`, the Chromatic tests didn't run :)
But the more I think about it, we should probably be running Chromatic even if root-level files are changed, since we don't want to introduce breaking builds.

ryan-roemer (approved):
🎉

PR #2469, opened 2022-09-21
use different GitHub token for Version Packages

### LINKED ISSUE #2434: "Version Packages" PRs don't run the PR checks
When a "Version Packages" PR is created (by the bot), the PR checks don’t run, so the PR can't be merged.

The current workaround is to hit "Close PR" and then "Reopen PR".

paulmarsicloud:
To help resolve #2434 - I believe this [thread](https://github.com/changesets/action/issues/187) and [solution](https://github.com/changesets/action/issues/187#issuecomment-1228413850) outline a potential temporary fix.

A token was generated (details to be provided in Slack) and this has been setup in the `release.yml` workflow.

Hopefully this works and correctly runs the workflows as the `victory-ci` user without us needing to close and re-open the PR!

We will need to test this properly after a merge to review if the "Version Packages" PR that is opened runs the CI tests/suite as expected. If it does, we do not need to revert this change. If it does not, then this change has not been worthwhile in fixing this bug

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+          token: ${{ secrets.VERSION_PACKAGES_TOKEN }}
 
       - name: Use Node.js
         uses: actions/setup-node@v2
@@ -54,5 +55,5 @@ jobs:
           version: pnpm run version
           publish: pnpm run publish
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.VERSION_PACKAGES_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,5 +55,7 @@ jobs:
           version: pnpm run version
           publish: pnpm run publish
         env:
-          GITHUB_TOKEN: ${{ secrets.VERSION_PACKAGES_TOKEN }}
+          # Note: we are using a different GITHUB_TOKEN due to the issues in this thread:
+          # https://github.com/changesets/action/issues/187
+          GITHUB_TOKEN: ${{ secrets.CHANGESETS_GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.changeset/breezy-berries-fix.md
+++ b/.changeset/breezy-berries-fix.md
@@ -1,0 +1,5 @@
+---
+"victory-zoom-container": patch
+---
+
+Test for Changesets GitHub Actions workflow fix


.github/workflows/release.yml
```diff
@@ -54,5 +55,5 @@ jobs:
           version: pnpm run version
           publish: pnpm run publish
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.VERSION_PACKAGES_TOKEN }}
```
ryan-roemer, line 58: Can you just add a YAML comment with links to what you posted in this PR description saying why we need this?

ryan-roemer (approved):
Nice find! One small comment suggestion, rest looks great!

I know @scottrippey is going to be disappointed he doesn't get to repeatedly close/re-open Version Packages PRs anymore 😛

scottrippey (approved):
Looks good!  Can you add some instructions on how we'll verify this works, post-merge?

.github/workflows/release.yml
scottrippey, line 58: I was thinking, could we name this:
```suggestion
          GITHUB_TOKEN: ${{ secrets.CHANGESETS_GITHUB_TOKEN }}
```

scottrippey (changes requested):

.github/workflows/release.yml
scottrippey: Interestingly, Spectacle does have a variable named `CHANGESETS_GITHUB_TOKEN` too, but I don't see it being used anywhere.
Spectacle doesn't have "Required PR Checks" enabled, so they can merge without running any checks.

paulmarsicloud:
> Looks good! Can you add some instructions on how we'll verify this works, post-merge?

Will add some instructions to the PR!

.github/workflows/release.yml
paulmarsicloud: agree with the name change - will change this up!

.github/workflows/release.yml
```diff
@@ -54,5 +55,7 @@ jobs:
           version: pnpm run version
           publish: pnpm run publish
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Note: we are using a different GITHUB_TOKEN due to the issues in this thread:
+          # https://github.com/changesets/action/issues/187
+          GITHUB_TOKEN: ${{ secrets.CHANGESETS_GITHUB_TOKEN }}
```
scottrippey, line 60: Looks great!
For testing sake, let's add a changeset to this PR (run pnpm changeset)
We don't necessarily have to publish it, we just want to see if it creates the PR and runs the checks.

paulmarsicloud: added this for testing: [.changeset/breezy-berries-fix.md](../blob/chore/version-packages-pr-fix/.changeset/breezy-berries-fix.md) 🤞

scottrippey (approved):

Issue #2478, opened 2022-09-25
Missing react dependency for victory-native v36


ruairidhgallagher:
# Questions
N/A

# Bug Reports

### Checklist

- [x] I have read through the [FAQ](https://formidable.com/open-source/victory/docs/faq) and [Guides](https://formidable.com/open-source/victory/guides/)

- [x] I am using the latest version of Victory

- [x] I've searched open issues to make sure I'm not opening a duplicate issue

### The Problem

Using Intel Macbook, 
within a React app, 
using react v17.0.2,
using yarn v3.2.3 
running **yarn install for victory-native v36.6.7** 
within the console I see the warnings below.

➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (p35168), requested by victory-core
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (p0e13c), requested by victory-axis
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (p31079), requested by victory-polar-axis
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (pdb675), requested by victory-shared-events
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (p9ca13), requested by victory-brush-container
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (p56173), requested by victory-cursor-container
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (pf3f25), requested by victory-selection-container
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (pc61c5), requested by victory-voronoi-container
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (pa4fd1), requested by victory-zoom-container
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (p6d809), requested by victory-bar
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (pd0f66), requested by victory-tooltip
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (pc8eb9), requested by victory-area
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (pbf46f), requested by victory-box-plot
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (p1dee1), requested by victory-brush-line
➤ YN0002: │ victory-native@npm:36.6.7 doesn't provide react (p7f754), requested by victory-candlestick
[...]

PR #2479, opened 2022-09-26
Chore: added `react`  as a peer dependency to `react-native`

scottrippey:
# What

- Adds `react` as a peerDependency for `react-native`
- Addresses #2478


gksander (approved):
LGTM

Issue #2449, opened 2022-08-30
events is not defined in <VictoryStack> props.


dlehddnjs:
**Describe the bug**
`events` is not defined in `<VictoryStack>` props.
It occurs type error in Typescript.

> TS2322: Type '{ children: Element[]; width: number; height: number; padding: { top: number; bottom: number; left: number; right: number; }; colorScale: string[]; events: { target: string; eventHandlers: { onPressIn: ({}: {}) => { ...; }[]; onPressOut: ({}: {}) => { ...; }[]; }; }[]; }' is not assignable to type 'IntrinsicAttributes & Pick<InferProps<{ bins: Requireable<number | (number | Date | null | undefined)[]>; categories: Requireable<(string | null | undefined)[] | InferProps<...>>; ... 26 more ...; width?: Validator<...> | undefined; }>, "height" | ... 22 more ... | "singleQuadrantDomainPadding"> & Partial<...> & Par...'.   Property 'events' does not exist on type 'IntrinsicAttributes & Pick<InferProps<{ bins: Requireable<number | (number | Date | null | undefined)[]>; categories: Requireable<(string | null | undefined)[] | InferProps<...>>; ... 26 more ...; width?: Validator<...> | undefined; }>, "height" | ... 22 more ... | "singleQuadrantDomainPadding"> & Partial<...> & Par...'.

**Victory version**
7.20.3

**Code Sandbox link**
https://snack.expo.dev/@simplefunction/victorystackevents
(It cannot be reproduced in snack, only occurs in IDE)

**To Reproduce**
<VictoryStack
            events={[
              {
                target: 'data',

**Expected behavior**
There should be no TS2322 error.

PR #2480, opened 2022-09-26
Typescript: victory-stack props

scottrippey:
# What
- Adds strong types to `victory-stack` props
- Addresses #2449


gksander (approved):
🎈

Issue #2475, opened 2022-09-23
VictoryTooltip width on Windows


cagha:
**Describe the bug**
Having a weird issue with tooltip padding.
This issue is happening on all browsers on Windows machines but not at all on Macs.
Top/Bottom padding works normal but Left/Right seems to be ignored on Windows.
While checking open bugs I noticed that the issue already appears in previously created sandboxes.. like this one for example: https://codesandbox.io/s/victory-brush-zoom-chart-forked-64tdfw

Windows:
<img width="699" alt="CleanShot 2022-09-23 at 15 51 49@2x" src="https://user-images.githubusercontent.com/114258949/192066820-5c24c3c1-7066-4841-989b-bb516e8cec82.png">

Mac:
<img width="654" alt="CleanShot 2022-09-23 at 15 52 04@2x" src="https://user-images.githubusercontent.com/114258949/192066858-a0016285-1c13-4581-a3a2-2b579f6f82c1.png">

**Victory version**
36.6.7

scottrippey:
With #2505 merged, can you check if this problem has been solved with the latest version?

PR #2505, opened 2022-10-19
Fix inaccurate text size calculations

EthanRutherford:
Addresses https://github.com/FormidableLabs/victory/issues/2475

As mentioned in the issue, the tooltip can appear to not have accurate sizing/padding in some cases.
I investigated and found that the tooltip tries to approximate the size that the text content will be, and then uses that to then size the tooltip box around the text. The logic for this size approximation lives [here](https://github.com/FormidableLabs/victory/blob/main/packages/victory-core/src/victory-util/textsize.ts).

The logic for approximating the text size involves reading from a list of hard-coded font metadata, and computing styles and glyph widths to reach a final width and height. (`victory-tooltip` then adds the padding, and uses that for the size of the tooltip).
There's a few issues with this, the primary being that if a font is being used which is not in this list, the measurement could be off by an arbitrary amount. While that's not the case here, it is of course possible for a third party consumer to use their own font-set which is not in this list. 

In this case, I discovered the culprit was the font-family *list*. It's common to list several fallback fonts for an element in css, but the logic here simply uses the first font family it finds in the list which it *has metadata for*. This may not match the font that is actually used during render, as the client may not have that font available. I am assuming that the reason this method doesn't make any attempt to query the browser for available fonts/measurements because this code may be called server-side, where it does not have access to the client it will be rendered on.

To remedy this issue, I've added logic to the tooltip component which measures the true size of the tooltip text using the DOM api, and does an immediate rerender with the accurate sizes. The rerender happens synchronously, before the browser does the next paint, and so the end user never sees the incorrect size.

before: 
image

after: 
image

### DIFF
--- a/.changeset/pretty-moles-laugh.md
+++ b/.changeset/pretty-moles-laugh.md
@@ -1,0 +1,5 @@
+---
+"victory-core": patch
+---
+
+Fixes #2475: Improve accuracy of text size measurements
--- a/.changeset/pretty-moles-laugh.md
+++ b/.changeset/pretty-moles-laugh.md
@@ -2,4 +2,4 @@
 "victory-core": patch
 ---
 
-Fixes #2475: Improve accuracy of text size measurements
+Improve accuracy of text size measurements (fixes #2475)
--- a/packages/victory-core/src/victory-util/textsize.ts
+++ b/packages/victory-core/src/victory-util/textsize.ts
@@ -283,7 +283,15 @@ const _getMeasurementContainer = memoize(() => {
   return containerElement;
 });
 
-const _measureDimensionsInternal = (
+const styleToKeyComponent = (style) => {
+  if (!style) {
+    return "null";
+  }
+
+  return `${style.angle}:${style.fontFamily}:${style.fontSize}:${style.letterSpacing}:${style.lineHeight}`;
+};
+
+const _measureDimensionsInternal = memoize((
   text: string | string[],
   style?: TextSizeStyleInterface,
 ) => {
@@ -316,11 +324,14 @@ const _measureDimensionsInternal = (
   containerElement.innerHTML = "";
 
   return { width, height };
-};
+}, (text, style) => {
+  const totalText = Array.isArray(text) ? text.join() : text;
+  const totalStyle = Array.isArray(style) ? style.map(styleToKeyComponent).join() : styleToKeyComponent(style);
+  return `${totalText}::${totalStyle}`;
+});
 
 export interface TextSizeStyleInterface {
   angle?: number;
-  characterConstant?: string;
   fontFamily?: string;
   fontSize?: number | string;
   letterSpacing?: string;
--- a/packages/victory-core/src/victory-util/textsize.ts
+++ b/packages/victory-core/src/victory-util/textsize.ts
@@ -290,6 +290,7 @@ const _measureDimensionsInternal = (
   const containerElement = _getMeasurementContainer();
 
   const lines = _splitToLines(text);
+  let heightAcc = 0;
   for (const [i, line] of lines.entries()) {
     const textElement = document.createElementNS(
       "http://www.w3.org/2000/svg",
@@ -303,6 +304,9 @@ const _measureDimensionsInternal = (
     textElement.style.fontFamily = params.fontFamily;
     textElement.style.letterSpacing = params.letterSpacing;
     textElement.textContent = line;
+    textElement.setAttribute("x", "0");
+    textElement.setAttribute("y", `${heightAcc}`);
+    heightAcc += params.lineHeight * params.fontSize;
 
     containerElement.appendChild(textElement);
   }
--- a/packages/victory-core/src/victory-util/textsize.ts
+++ b/packages/victory-core/src/victory-util/textsize.ts
@@ -325,7 +325,11 @@ export interface TextSizeStyleInterface {
 
 // Stubbable implementation.
 export const _approximateTextSizeInternal = {
-  impl: (text: string | string[], style?: TextSizeStyleInterface) => {
+  impl: (
+    text: string | string[],
+    style?: TextSizeStyleInterface,
+    __debugForceApproximate = false,
+  ) => {
     // Attempt to first measure the element in DOM. If there is no DOM, fallback
     // to the less accurate approximation algorithm.
     const isClient =
@@ -333,7 +337,7 @@ export const _approximateTextSizeInternal = {
       typeof window.document !== "undefined" &&
       typeof window.document.createElement !== "undefined";
 
-    if (!isClient) {
+    if (!isClient || __debugForceApproximate) {
       return _approximateDimensionsInternal(text, style);
     }
 
--- a/packages/victory-core/src/victory-util/textsize.ts
+++ b/packages/victory-core/src/victory-util/textsize.ts
@@ -291,44 +291,46 @@ const styleToKeyComponent = (style) => {
   return `${style.angle}:${style.fontFamily}:${style.fontSize}:${style.letterSpacing}:${style.lineHeight}`;
 };
 
-const _measureDimensionsInternal = memoize((
-  text: string | string[],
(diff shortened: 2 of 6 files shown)


scottrippey (commented):
A great approach, and I would approve, but there's actually a PR already open that does the same thing! #2352
Let's see if the PR above fixes the issue too? 
This is a good reminder, too, that we should try to get one of these PRs merged!

EthanRutherford:
@scottrippey oh wow, whoops 😅 well, if there's a way I can help get one of these two merged let me know!

EthanRutherford:
@scottrippey I've updated the PR, bringing in the logic from the older PR and fixing some issues.

ryan-roemer:
Thanks for this work! Can we get a [changeset](https://github.com/FormidableLabs/victory/blob/main/CONTRIBUTING.md#add-a-changeset) added for this change too?

.changeset/pretty-moles-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+"victory-core": patch
+---
+
+Fixes #2475: Improve accuracy of text size measurements
```
ryan-roemer, line 5: The changeset changelog plugin we use ([@svitejs/changesets-changelog-github-compact](https://github.com/svitejs/changesets-changelog-github-compact/blob/main/packages/changesets-changelog-github-compact/README.md)) is a little bit picky about magic links and stuff. Here's what I'd suggest for final changelog correctly linking: 

```suggestion
Improve accuracy of text size measurements (fixes #2475)
```

scottrippey (approved):
Nice work, looks great!

stories/victory-label.stories.js
```diff
@@ -415,6 +415,30 @@ export const LineHeight = () => {
...
+      />
+      <VictoryScatter
+        {...defaultScatterProps}
+        labelComponent={
+          <VictoryLabel
+            lineHeight={[2, 1, 3]}
+            text={[
+              "اختبار اللغات التي تُقرأ من اليمين إلى اليسار",
+              "مثل العربية",
+              "هناك أكثر من ذلك بكثير",
```
scottrippey, line 436: Looks like something is going wrong here:
https://snapshots.chromatic.com/snapshots/5b4acf7c54c0490024d5980b-6351bb5c8bc2046ba030855b/capture.png

EthanRutherford: Hmm.. I'll take another look, but it *was* working on my machine when I added these. I'll try and figure out what's wrong there.

EthanRutherford: Yeah, I'm not sure.. It looks just fine locally:
image

EthanRutherford: In fact, if I visit https://5b4acf7c54c0490024d5980b-cxfpozovry.chromatic.com/?path=/story/victorylabel--line-height it also looks correct. Strange...

scottrippey: It looks fine on my phone too. But that image above is from Chromatic's screenshots, so I'm not sure what they're doing differently. If we can't reproduce, then we might just want to accept it, because I'm sure it's better than before.

EthanRutherford: Yeah, I'm not having any luck reproducing. What do we know about the environment that chromatic uses to render?

EthanRutherford: Looks like chromatic was using chrome, so.. I'm not sure what's the deal there.

EthanRutherford: @scottrippey I believe we ought to just go ahead and merge, I've not managed to find a repro.

scottrippey: Just as a follow-up to this thread:
I was able to reproduce in Safari, so maybe Chromatic is using WebKit too?
Either way, in Safari, the text-layout issue is not new, and I get the same layout in the `main` branch too.  
So, I think we should merge, and address the text-layout issue separately.

packages/victory-core/src/victory-util/textsize.ts
```diff
@@ -243,6 +243,81 @@ const _approximateTextHeightInternal = (text: string | string[], style) => {
...
+  element.style.position = "fixed";
+  element.style.top = "-9999px";
+  element.style.left = "-9999px";
+
+  document.body.appendChild(element);
+
+  return containerElement;
+});
+
+const _measureDimensionsInternal = (
```
scottrippey, line 286: I'd like to bring up performance...
Currently, it looks like we are calling `approximateTextSize` inside the render method of `VictoryLabel`.  This could certainly be problematic, considering we're doing DOM manipulation in the method now.  And every re-render (eg. animation) will have to execute all this logic repeatedly, even though the dimensions are unlikely to change.

I'd like to improve this approach, if possible.  Some ideas that come to mind:
1. On initial render, use the old `approximateTextSize` (using pre-defined glyph measurements), then inside a `useLayoutEffect` we could perform these measurements, to get a higher accuracy
2. We should memoize these calculations somewhere.  Probably using `useMemo` inside `VictoryLabel`
3. Is the `approximateTextSize` logic able to know whether it's accurate?  Like, if we're measuring unicode symbols that aren't defined, or if the font face is undefined, then we know that the "approximation" will be wrong?   If so, we could ONLY perform the "measure" logic if we think the approximation is wrong.  Not sure if it's totally worth it, but something to think about.

stories/victory-label.stories.js
EthanRutherford: Ok, that makes sense. I'm on windows, so I unfortunately don't have access to test in safari.

packages/victory-core/src/victory-util/textsize.ts
EthanRutherford: I can certainly look into avoiding expensive re-calculations. `useMemo` is likely to be our best bet. 

Since `useLayoutEffect` blocks render until the layout effect has been applied, where we'd then do a setState and cause a rerender, this idea would pretty much just make the initial render even slower.

I'll do some investigating and measuring and see if there's anything clever I can do.

scottrippey: Yeah `useLayoutEffect` is probably the wrong place, maybe `useEffect` but that could cause a CLS.  
I guess that's why I was thinking "#3" above would help, so that we could avoid measurements completely, for most common scenarios.

EthanRutherford: Unfortunately, I don't think 3 is going to be reliable. It's possible to get false positives, where we think we know we have the correct font-family etc. but when we render, the used font differs from the one we though would be used. I mention as much in the PR description, but essentially there are various ways that the font-family we're using for our computation would be different than the rendered font.

I have added memoization to the measurement function, so repeated measurements of the same label/tooltip/whatever text now comes for free.

packages/victory-core/src/victory-util/textsize.ts
```diff
@@ -316,11 +324,14 @@ const _measureDimensionsInternal = (
   containerElement.innerHTML = "";
 
   return { width, height };
-};
+}, (text, style) => {
+  const totalText = Array.isArray(text) ? text.join() : text;
```
scottrippey: This is cool! 
Maybe you could really simplify by doing `JSON.stringify` here, since it's ultra optimized.

scottrippey (approved):

packages/victory-core/src/victory-util/textsize.ts
scottrippey: Ok, sounds good. 
I looked at the "approximate" logic and yes, it's hard to improve it. The algorithm that the browser uses to render text is impossible to recreate in user land. Measuring is the only reliable way, even though it's quite expensive.

packages/victory-core/src/victory-util/textsize.ts
EthanRutherford: I chose not to go for json.stringify simply because I wanted to make sure the attributes were in a consistent order, but that may or may not actually be a concern we're worried about. Happy to change it if we feel it's not a big deal.

PR #2535, opened 2022-12-13
Chore: Migrate victory-tooltip to TypeScript

gwynethrose:
- Migrated component files and index to TypeScript.
- Removed old .d.ts file, which appeared to be outdated.


packages/victory-tooltip/src/victory-tooltip.tsx
```diff
@@ -159,7 +146,7 @@ export default class VictoryTooltip extends React.Component {
   getPolarOrientation(props, datum) {
```
gwynethrose, line 146: I didn't type most of these helper functions, since doing so mostly led to extraneous null checks or would require creating a new type to represent the evaluated, defaulted props, which didn't appear to be a pattern elsewhere.

sutt0n (approved):

Issue #2489, opened 2022-10-06
`Scale` prop missing from VictoryAxis propTypes


jessicagomesng:
**Describe the bug**
Scale prop missing from `VictoryAxis` propTypes declaration. This is causing our console to log out an error about the prop, but it does appear from the documentation that this prop can be applied to VictoryAxis. 

**Victory version**
36.0.0

**Expected behavior**
Scale prop should not throw an error when used in conjunction with VictoryAxis. Our code is something like: 

`<VictoryAxis scale={{y: 'linear'}} />`

**Screenshots**
![Screen Shot 2022-10-07 at 10 46 00 AM](https://user-images.githubusercontent.com/61810279/194427412-5ff1d69a-28b1-4959-b4e9-3bc8eaf9af7c.png)

**Desktop (please complete the following information):**
 - OS: OS
 - Browser : Chrome

docmars:
I'm also seeing this on `VictoryChart`, where it should (and is) a supported prop that's working fine during runtime.

```
Warning: Failed prop type: Invalid prop `scale` supplied to `VictoryChart`.
```

galgord:
I am seeing a similar issue with the domain property using victory native and victory area
Warning: Failed prop type: Invalid prop `domain` supplied to `VictoryArea`.

Decezaris:
I had the same problem here with `VictoryChart`, what solved for me was put `x` and `y` inside scale prop as bellow 
`scale={{ x: 'time', y: 'linear' }}`

MosefAsad:
> 

This issue is also happening with `VictoryGroup` and adding the `y` prop helped the error go away.

SaeedZhiany:
It's still an error in `v36.6.10`.

```console
Failed prop type: Invalid prop `scale` supplied to `VictoryChartimpl`.
```

```tsx
<VictoryChart
  key={'victory chart'}
  {/* other props */}
  scale={{x: 'time'}}
>
</VictoryChart>
```

@becca-bailey 
According to these types, it should be OK, it's kinda weird why we are getting the warning.

<img width="592" alt="image" src="https://github.com/FormidableLabs/victory/assets/13766432/1d10a366-b386-4f4b-973d-69dc1be2b837">

<img width="485" alt="image" src="https://github.com/FormidableLabs/victory/assets/13766432/e933068a-b5dd-4482-89d3-1cc09fb907e7">

katche70:
@SaeedZhiany 
[...]

PR #2537, opened 2022-12-15
Update PropTypes to support single axis scale values

scottqmn:
Related issue #2489 

Users would get PropType errors on `scale` props when single axis values were provided (e.g. `scale={{ x: 'log' }}`) even though documentation and demos both have examples of single axis scale props and the components functioned normally. 

PropTypes for `scale` have been updated to no longer require both axises. `PropTypes.shape` was replaced with `PropTypes.exact` in order to catch invalid properties (e.g. `{ z: 'log' }`)

### Caveat
TypeScript error is suppressed on updated prop types. The error appears when `isRequired` is removed from either axis properties. A verbose workaround could be to list each possible combination of axis properties to keep `isRequired` like so:
```
scale: PropTypes.oneOfType([
  CustomPropTypes.scale,
  PropTypes.exact({
    x: CustomPropTypes.scale.isRequired,
    y: CustomPropTypes.scale.isRequired,
  }),
  PropTypes.exact({
    x: CustomPropTypes.scale.isRequired,
  }),
  PropTypes.exact({
    y: CustomPropTypes.scale.isRequired,
  }),
])
```

### DIFF
--- a/packages/victory-core/src/victory-util/common-props.tsx
+++ b/packages/victory-core/src/victory-util/common-props.tsx
@@ -228,11 +228,12 @@ const baseProps: React.WeakValidationMap<VictoryCommonProps> = {
       y: CustomPropTypes.domain.isRequired,
     }),
   ]),
+  // @ts-expect-error TODO: synchronize the type with this PropTypes
   scale: PropTypes.oneOfType([
     CustomPropTypes.scale,
-    PropTypes.shape({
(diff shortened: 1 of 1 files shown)


docmars:
Hey there! Any momentum on this? I'd very much like to stop seeing the console error for this 😁 

Thanks for getting this started!

PR #2547, opened 2023-01-02
fix: fix using React.ComponentProps with VictoryLine component

Waltari10:
Looking at file `myApp/node_modules/victory-line/lib/victory-line.d.ts` file the built type file ended up with a constructor that takes props as type of any:
```
declare class VictoryLineBase extends React.Component<VictoryLineProps> {
    constructor(props: any); // Breaks typing
    static animationWhitelist: string[];
```

When trying to parse the components props one ended up with `any` type for `VictoryLineProps`. However inferring types for other components like `VictoryScatter` worked fine because it didn't have constructor defined in code.

```
import {
  VictoryLine,
  VictoryScatter,
} from "victory-native";

type VictoryScatterProps = React.ComponentProps<typeof VictoryScatter>; // Correct typing
type VictoryLineProps = React.ComponentProps<typeof VictoryLine>; // any type
```

So in this PR just removing the constructor to avoid it getting any type. Doesn't look like the constructor is needed or used for anything anyway.

### DIFF
--- a/.changeset/neat-trains-doubt.md
+++ b/.changeset/neat-trains-doubt.md
@@ -1,0 +1,5 @@
+---
+"victory-line": minor
+---
+
+Fix using React.ComponentProps with VictoryLine component
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "author": "Formidable",
   "license": "MIT",
   "bugs": {
-    "url": "https://github.com/formidablelabs/victory/issue"
+    "url": "https://github.com/formidablelabs/victory/issues"
   },
   "homepage": "https://formidable.com/open-source/victory",
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   "author": "Formidable",
   "license": "MIT",
   "bugs": {
-    "url": "https://github.com/formidablelabs/victory/issues"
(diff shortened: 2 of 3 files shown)


Waltari10:
Hi, is it possible to get some eyeballs on this PR?.. Would be great to get rid of a patch-package in our codebase 😀 

@ryan-roemer @becca-bailey ?

ryan-roemer:
@Waltari10 Thanks for this! We'll try and get around to this next week if we can. I'm going to add a review to get CI running and checking.

ryan-roemer (commented):
.

ryan-roemer (dismissed):
CI

ryan-roemer:
@scottrippey @gksander Any idea why this CI is stuck in `Expected — Waiting for status to be reported`? This is a forked PR, but I don't think that should result in this particular situation... ?

gksander:
I'm going to try closing and reopening this PR to see if it'll trigger GitHub Actions to send another approval request so we can get CI ran on this!

gksander:
We're going to have to look into why forked PRs aren't triggering approval requests for CI runs, but in the meantime – I went ahead and just pushed up a couple dummy commits to the branch to trigger the workflows.

PR #2667, opened 2023-10-18
Upgrade CI to Node 18 and deprecate Node 14

carbonrobot:
Upgrades the CI to default to Node 18.x and refactors the common CI setup into a reusable action.

- Removes CI run for Node 14.x
- Upgrades actions/checkout to v4
- Upgrades actions/setup-node to v3


gksander:
Completed just fine locally, so looks like we've got a little bit of cache investigation to undertake (and I don't think GH offers a way to e.g. "nuke all 1367 cache artifacts"), ought to be a fun little adventure.

carloskelly13 (approved):
Nice! Love the streamlining of the CI set up.

PR #2670, opened 2023-11-02
Publish aggregate release notes to Github

### LINKED ISSUE #2408: Infra: Create GitHub Releases on npm release
We now only have single tags for releases with our custom git tagging strategy. It would be nice to have real GH releases again.

Options:

- [ ] See if https://github.com/changesets/action/pull/193 for changesets (adding `createGithubReleases`) works for us with our custom git tagging scheme.
- [ ] Do it ourselves with maybe a separate action keyed on a git tag happening using something like https://github.com/softprops/action-gh-release
    - [ ] We'll need to ingest changelogs probably in a custom script to generate release notes.

carbonrobot:
Publish aggregate release notes to Github

Uses modified code from [this fork](https://github.com/dotansimha/changesets-action) to allow aggregate release notes publishing. This can be removed when this is a [fully supported feature in changesets](https://github.com/changesets/action/pull/193).

Closes #2408


scripts/release.ts
```diff
@@ -0,0 +1,120 @@
...
+    ast.children = ast.children.slice(headingStartInfo.index + 1, endIndex);
+  }
+  return {
+    content: unified().use(remarkStringify).stringify(ast),
+    highestLevel,
+  };
+}
+
+const createAggregatedRelease = async () => {
+  const octokit = new Octokit({ auth: process.env.CHANGESETS_GITHUB_TOKEN });
```
ryan-roemer, line 78: Isn't the GH action setting `GITHUB_TOKEN` on `env`?

.github/workflows/release.yml
```diff
@@ -41,3 +41,12 @@ jobs:
           # https://github.com/changesets/action/issues/187
           GITHUB_TOKEN: ${{ secrets.CHANGESETS_GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      # Note: We run a custom release notes script so we can generate a single aggregate
+      # release notes file for all packages. This can be removed when `changesets` supports
+      # this feature natively.
+      - name: Github Release notes
+        if: steps.changesets.outputs.published == 'true'
+        run: npx ts-node ./scripts/release-notes.ts
```
ryan-roemer, line 50: I've personally found `npx` super slow, not sure if it's gotten better... Given `ts-node` is relatively small and we're adding a number of deps just for this script, should we add `ts-node` to root level `package.json:devDependencies`?

ryan-roemer (approved):
Looks great! One totally optional comment, and other comment I think needs a fix. Thanks so much for this!

scripts/release.ts
carbonrobot: Victory uses a custom github token due to the following

https://github.com/FormidableLabs/victory/blob/main/.github/workflows/release.yml#L34-L36

.github/workflows/release.yml
carbonrobot: Good shout. Will do

carbonrobot: Updated

PR #2671, opened 2023-11-07
fix: allow empty tick values

msabree:
Issue Link: https://github.com/FormidableLabs/victory/issues/1300

Allow empty tickValues and display an empty graph. The suggested work around was `tickValues={[null]}`. We can now do: `tickValues={[]}`.

Before with empty tick values:

Screenshot 2023-11-07 at 2 00 03 PM

After:

Screenshot 2023-11-08 at 9 56 24 AM


packages/victory-axis/src/helper-methods.tsx
```diff
@@ -595,7 +595,8 @@ export const getBaseProps = (props, fallbackProps) => {
         ? { [otherAxis]: props.scale[otherAxis] }
         : undefined,
   };
-  return ticks.reduce((childProps, tickValue, index) => {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+  return (ticks as number[]).reduce((childProps, tickValue, index) => {
```
scottrippey, line 599: This seems weird, is this change necessary?

scottrippey (approved):
Tiny suggestion, otherwise lgtm

demo/ts/components/victory-axis-demo.tsx
```diff
@@ -269,6 +269,14 @@ export default class VictoryAxisDemo extends React.Component<
...
             ]}
           />
+
+          <VictoryAxis
+            label="Empty Values"
+            padding={{ top: 40, bottom: 40, left: 40, right: 90 }}
+            orientation="right"
+            style={styleOverrides}
+            tickValues={[]}
+          />
```
scottrippey, line 279: We have a storybook for this repo too, which takes screenshots of all stories.  (Storybook is separate from the `demo` folder).
Can you please add this test-case to the Storybook story too?

packages/victory-axis/src/helper-methods.tsx
msabree: It was giving a type check error without it. I assume since i force [''] into what is expecting a number[]. I'm sure there may be a cleaner way to do it.

demo/ts/components/victory-axis-demo.tsx
msabree: will do

packages/victory-axis/src/helper-methods.tsx
scottrippey: The reason it's weird, though, is because of the need to disable `no-unnecessary-type-assertion` ... which is basically saying "it's already `number[]`, no need for `as number[]`.
Can you check what happens if you revert this change?  Maybe it was broken during development but now it can be changed back?
If it's still broken, then it's ok to merge.

msabree: Screenshot 2023-11-12 at 11 55 35 AM

stories/victory-axis.stories.js
```diff
@@ -133,6 +133,9 @@ export const TickValues = () => {
       <VictoryChart {...defaultChartProps} scale={{ x: "log" }}>
         <VictoryAxis tickValues={[1, 3, 5, 7, 10, 50, 100, 500, 1000]} />
       </VictoryChart>
+      <VictoryChart {...defaultChartProps}>
+        <VictoryAxis label={'Empty Tick Values'} tickValues={[]} orientation="right" />
+      </VictoryChart>
```
scottrippey, line 138: ❤️

scottrippey (approved):

stories/victory-axis.stories.js
scottrippey, line 1: Looks like CI checks are failing; error is buried, but looks like it's due to prettier failing on this file.

scottrippey (approved):

PR #2677, opened 2023-11-21
Update storybook and webpack

Cperez2187:
Update Storybook to v7 and latest webpack.

This updates seems to require us to update some of the babel setup we have including the root file name to `babel.config`. See this [issue](https://github.com/storybookjs/storybook/issues/22357) for reference on why this is needed


KenanYusuf:
Fixed in https://github.com/FormidableLabs/victory/pull/2743

PR #2679, opened 2023-11-27
converted all functional components to use default parameters instead of defaultProps

acharyakavita:
As per https://github.com/reactjs/rfcs/pull/107 all React functional components are converted to use default parameters instead of defaultProps.


packages/victory-area/src/area.tsx
```diff
@@ -153,7 +153,7 @@ export const Area: React.FC<AreaProps> = (props) => {
 
   const line = renderLine
     ? React.cloneElement(
-        pathComponent!,
+        pathComponent,
```
acharyakavita, line 156: Got a message saying ` This assertion is unnecessary since it does not change the type of the expression.eslint[@typescript-eslint/no-unnecessary-type-assertion](https://typescript-eslint.io/rules/no-unnecessary-type-assertion)`. so I had to remove it

PR #2681, opened 2023-12-19
Add a holiday animation example

carbonrobot:
Add a fun holiday example to the Gallery!

Also updates the opengraph metadata for Gallery pages so we can share these properly in Social networks

santa


carloskelly13:
This is great, but I'm seeing a bit of sizing problems with overflow on the gallery page.

Screenshot 2023-12-19 at 3 01 52 PM

carloskelly13:
I wonder if a quick fix is a poster screenshot of the animation.

carloskelly13 (approved):
Ho ho ho 🎄

PR #2691, opened 2024-01-04
Fix Uncaught TypeError: date.getDay is not a function in @d3fc/d3fc-discontinuous-scale/**/skipWeekends.js

bl5ck:
When combine `VictoryVoronoiContainer` with 
```
scaleDiscontinuous(
    d3Scale.scaleTime()
  ).discontinuityProvider(discontinuitySkipWeekends())
```
from [d3fc-discontinuous-scale](https://github.com/d3fc/d3fc/blob/master/packages/d3fc-discontinuous-scale/src/discontinuity/skipWeekends.js)
there would be an error: **TypeError: date.getDay is not a function** when hovering the Chart.


bl5ck:
I see. The sample data indeed uses Date instance for the X-axis. I guess there is no choice than convert it to timestamp and implement skip weekend function myself.

PR #2705, opened 2024-01-09
Use parallel builds in workflow for improved CI times

carbonrobot:
Improves the CI run time by running the 4 CI checks in parallel and only running specific Node versions for checks where that matters.

image


package.json
```diff
@@ -175,19 +173,6 @@
         "clean:cache:modules"
       ]
     },
-    "check": {
```
ryan-roemer, line 178: Depending on where we're heading with wireit (e.g. if we're not removing it) we might want to keep `check` (but fine to nuke `check:ci`) for developers to be able to run `pnpm check` in localdev.

ryan-roemer: Oh, and here are the existing instances of `pnpm run check` and `pnpm check` -- https://github.com/search?q=repo%3AFormidableLabs%2Fvictory+%28%22pnpm+run+check%22+OR+%22pnpm+check%22%29&type=code

carbonrobot: Restored check command

mattjw79 (approved):
LGTM

PR #2712, opened 2024-01-12
Migrate victory-legend to TypeScript

KenanYusuf:
- Migrated `victory-legend` package files to TypeScript
- Removed old .d.ts file


packages/victory-core/src/victory-util/add-events.tsx
```diff
@@ -68,9 +68,9 @@ export interface EventsMixinClass<TProps> {
     defaultAnimationWhitelist: string[],
   ): React.ReactElement;
   getComponentProps(
-    component: React.ReactElement,
+    component: React.ReactNode,
     type: string,
-    index: number,
+    index: string | number,
```
KenanYusuf, line 73: This can be a number, or a key of an object (string)

packages/victory-core/src/victory-util/add-events.tsx
```diff
@@ -68,9 +68,9 @@ export interface EventsMixinClass<TProps> {
     defaultAnimationWhitelist: string[],
   ): React.ReactElement;
   getComponentProps(
-    component: React.ReactElement,
+    component: React.ReactNode,
```
KenanYusuf, line 71: Used a more inclusive type for the component, as it can also be a `string` or `string[]`

carbonrobot (approved):

PR #2722, opened 2024-01-17
Refactor reassigned props

KenanYusuf:
There are a bunch of cases where we are reassigning component properties directly through the props (or equivalent) param. This PR ensures that props are assigned to a const rather than mutating the original object.

**Note:** There are _lots_ of cases where we're reassigning other, non-prop, params. Following this PR, I will create another which adds the [no-param-reassign](https://eslint.org/docs/latest/rules/no-param-reassign) eslint rule and refactors the remaining cases.


carbonrobot (approved):
Not all heroes wear capes!

PR #2723, opened 2024-01-17
Migrate line test to typescript

carbonrobot:
Migrate line test to typescript

### DIFF
--- a/packages/victory-line/src/victory-line.test.tsx
+++ b/packages/victory-line/src/victory-line.test.tsx
@@ -194,9 +194,9 @@ describe("components/victory-line", () => {
       );
 
       const line = container.querySelector("path");
-      const renderedData = JSON.parse(
-        line!.getAttribute("data-json")!,
-      ).map(({ t }) => t);
+      const renderedData = JSON.parse(line!.getAttribute("data-json")!).map(
+        ({ t }) => t,
+      );
 
       expect(renderedData).toEqual([0, 1]);
     });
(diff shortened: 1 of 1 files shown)


packages/victory-line/src/victory-line.test.tsx
```diff
@@ -145,15 +146,6 @@ describe("components/victory-line", () => {
...
 
-    it("renders data values with null accessor", () => {
-      const data = [1, 2, 3, 4];
-      const { container } = render(
-        <VictoryLine data={data} x={null} y={null} />,
-      );
-      const lines = container.querySelectorAll("path");
-      expect(lines).toHaveLength(1);
-    });
-
```
carbonrobot, line 156: This test is not valid for our type definitions, and adjusting the type definitions would require us to change many components in a cascade bc of dependencies.

packages/victory-line/src/victory-line.test.tsx
```diff
@@ -322,11 +314,11 @@ describe("components/victory-line", () => {
       );
       const path = container.querySelector("path");
 
-      expect(path.getAttribute("aria-label")).toEqual(
+      expect(path?.getAttribute("aria-label")).toEqual(
         `data point 3's x value is 2`,
       );
 
-      expect(parseInt(path.getAttribute("tabindex"))).toEqual(3);
+      expect(parseInt(path?.getAttribute("tabindex") || "")).toEqual(3);
```
scottrippey, line 321: Conversion looks good!  Merge away.

Possible improvement: switch to `!` instead of `?` when in tests, because error messages would be better.  Like `cannot access 'getAttribute' of null` versus `expected "" to equal 3`

carbonrobot: done

scottrippey (approved):

PR #2777, opened 2024-02-05
Refactor victory-accessible-group to a function component

KenanYusuf:
Refactor victory-accessible-group to a function component


Burnett2k:
Looks good 🚀 . 
I tried to test this on the PR environment by viewing the docs, but it looks like we don't have a playground link / example there in the docs. Not needed for this PR, but having a working example in there could be helpful. It appears we have working examples for most other items such as VictoryLabel
image

Issue #2784, opened 2024-02-06
TypeScript errors since 36.8.6 version


TomCorvus:
**Describe the bug**
Since 36.8.6 version, I have TypeScript errors on multiple props with victory-native components.

**Victory version**
36.8.6 or 36.9.0

**Code Sandbox link**
```
<VictoryLine
	data={data}
	name="line1"
	interpolation="monotoneX"
	style={{
		labels: { fill: palette.bubblegum.bubblegum50 },
		data: { stroke: palette.bubblegum.bubblegum50 },
	}}
/>;
```
OR
```
<VictoryVoronoiContainer
	onTouchStart={() => onTouchStartHandler()}
	onTouchEnd={() => onTouchEndHandler()}
	voronoiBlacklist={['scatter1', 'scatter2']}
	voronoiDimension="x"
/>;
```

**To Reproduce**
In code above, `onTouchStart`, `onTouchEnd`, `data`, `name`, `interpolation`, `style` generate TS errors.
I didn't check others charts.

**Expected behavior**
No TS errors.

**Screenshots**
Not applicable

**Additional context**
TypeScript `^5.3.3`

PR #2785, opened 2024-02-06
Victory native type fixes

KenanYusuf:
# Problem

Relates to #2784

There were a couple of issues with the types in the `victory-native` package, caused by #2739.

## Wrapped components

Components that were wrapped with the `wrapCoreComponent` HOC were _only_ accepting props that were passed to the defaultProps object. For example:
```tsx
export const VictoryLine = wrapCoreComponent({
  Component: VictoryLineBase,
  defaultProps: {
    width: Dimensions.get("window").width,
  },
});

const Example = () => (
  <VictoryLine
    width={400} // ✅ exists in defaultProps
    data={[]} // ❌ does not exist in defaultProps
  />
);
```

I have updated this by ensuring that the base component's props are passed as a generic to the `wrapCoreComponent` function:
```tsx
export const VictoryLine = wrapCoreComponent<VictoryLineProps>({
  Component: VictoryLineBase,
  defaultProps: {
    width: Dimensions.get("window").width,
  },
});

const Example = () => (
  <VictoryLine
    width={400} // ✅ exists in VictoryLineProps
    data={[]} // ✅ exists in VictoryLineProps
  />
);
```

## **Container** components

The `victory-*-container` component props were being typed incorrectly as the type of the _original_ component props. For example:
```tsx
const combinedMixin = flow(originalVoronoiMixin, nativeVoronoiMixin);

export const voronoiContainerMixin = (base): VictoryVoronoiContainerBase =>
  combinedMixin(base);

export const VictoryVoronoiContainer = voronoiContainerMixin(VictoryContainer)
```

In the above example, `VictoryVoronoiContainerBase` is the `typeof` the base victory component. To fix this, the types were updated so that the return type of `combinedMixin` is a class component with the props for the `victory-native` version:

```tsx
const combinedMixin: (
  base: React.ComponentClass,
) => React.ComponentClass<VictoryVoronoiContainerNativeProps> = flow(
  originalVoronoiMixin,
  nativeVoronoiMixin,
);

export const voronoiContainerMixin = (base: React.ComponentClass) =>
  combinedMixin(base);

export const VictoryVoronoiContainer = voronoiContainerMixin(VictoryContainer);
```

### DIFF
--- a/packages/victory-selection-container/src/victory-selection-container.tsx
+++ b/packages/victory-selection-container/src/victory-selection-container.tsx
@@ -120,4 +120,3 @@ export function selectionContainerMixin<
 
 export const VictorySelectionContainer =
   selectionContainerMixin(VictoryContainer);
-export type VictorySelectionContainer = typeof VictorySelectionContainer;
--- a/.changeset/short-ads-hug.md
+++ b/.changeset/short-ads-hug.md
@@ -1,0 +1,11 @@
+---
+"victory-brush-container": patch
+"victory-cursor-container": patch
+"victory-line": patch
+"victory-native": patch
+"victory-selection-container": patch
+"victory-voronoi-container": patch
+"victory-zoom-container": patch
+---
+
+Fix victory-native component prop types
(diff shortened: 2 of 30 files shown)


packages/victory-selection-container/src/victory-selection-container.tsx
```diff
@@ -120,4 +120,3 @@ export function selectionContainerMixin<
 
 export const VictorySelectionContainer =
   selectionContainerMixin(VictoryContainer);
-export type VictorySelectionContainer = typeof VictorySelectionContainer;
```
KenanYusuf, line 123: These types are no longer necessary for extending types in the `victory-native` package

carbonrobot (approved):

PR #2788, opened 2024-02-07
Convert victory-animation to function component

KenanYusuf:
Convert victory-animation to function component

## Before
https://github.com/FormidableLabs/victory/assets/9557798/1f425984-3806-4741-ba2e-0e327538f760

## After
https://github.com/FormidableLabs/victory/assets/9557798/7f80296c-9547-43ce-82e9-62b75c5c3a0e

### DIFF
--- a/.changeset/eleven-geckos-trade.md
+++ b/.changeset/eleven-geckos-trade.md
@@ -1,0 +1,5 @@
+---
+"victory-core": patch
+---
+
+Convert victory-animation to function component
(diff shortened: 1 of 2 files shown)


carbonrobot (changes requested):
Before we merge this, I'm adding some new storybook ui tests we can use as a comparison with main.

carbonrobot (approved):
Scratch that. We are able to use the Demo to confirm. Looks good to me!

PR #2789, opened 2024-02-07
Refs/heads/405 victorypie curved text labels

acharyakavita:
Victory pie curved text labels


docs/src/content/docs/victory-pie.md
```diff
@@ -271,6 +276,16 @@ The `labelPlacement` prop specifies the angular placement of each label relative
 />
 ```
 
+```playground
+<VictoryPie
+  data={sampleData}
+  labels={({ datum }) => `y: ${datum.y}`}
+  labelPlacement={"curved"}
```
carbonrobot, line 283: We can omit the data binding here

```suggestion
  labelPlacement="curved"
```

carbonrobot (commented):
This is a great start!

I think we should improve the API a little.

For all labelPlacement types, the value of labelPlacement affects the rendering of the labelComponent.

I think we want to stay with that pattern here and not introduce a new property for curvedLabels. Looking through your code, I think we can adapt it roughly by doing the following:

1. add a new property to `Label` called `radius`. This aligns with other props like `arc` that alter the rendering, allowing people to make a curved label regardless of the chart used.
2. when labelPlacement="curved", pass the radius value to the `labelComponent` property internally in `PieChart`
```
const labelProps = ...
if(labelComponent="curved"){
  labelProps.radius = pieSliceRadius;
}
```

PR #2796, opened 2024-02-09
Refactor jest configuration

carbonrobot:


ryan-roemer:
I've pulled the branch and I'm running locally.

Burnett2k (approved):
I was able to pull down the branch on my M1 Mac, run the commands listed in the description, and receive successes on all of them. All the jest tests passed as well.

KenanYusuf (commented):
I am getting the following error and test failures for the `victory-selection-container` and `victory-errorbar` packages:
```
    src/selection-helpers.test.tsx:7:28 - error TS2339: Property 'scaleLinear' does not exist on type 'typeof import(".../victory/packages/victory-selection-container/node_modules/victory-vendor/d3-scale")'.

    7 const scale = { x: d3Scale.scaleLinear(), y: d3Scale.scaleLinear() };
```

When I add `victory-vendor` as a dependency for both of those packages, it fixes the error and test failures.

PR #2804, opened 2024-02-13
Upgrade babel and use env presets for browsers

carbonrobot:
Upgrades all babel dependencies and uses the `@babel/preset-env` for configuring plugins.

[`@babel/preset-env`](https://babeljs.io/docs/babel-preset-env) uses [Browserslist](https://github.com/browserslist/browserslist) to specify target transforms. In our case, we are using [`> 0.5%, last 2 versions, Firefox ESR, not dead`](https://browsersl.ist/#q=%3E+0.5%25%2C+last+2+versions%2C+Firefox+ESR%2C+not+dead). This is the default recommended config from the community in order to promote best practices and avoid common issues. It targets 88.8% of browsers globally.

 This change reduces the file sizes dramatically

image

Attached is a zip comparison of the files
[diff.zip](https://github.com/FormidableLabs/victory/files/14272484/diff.zip)

Test package deployed to npm `victory@36.9.2-next.0`


package.json
```diff
@@ -124,6 +113,12 @@
       "@types/eslint": "8.4.3"
     }
   },
+  "browserslist": [
+    "> 0.5%",
+    "last 2 versions",
+    "Firefox ESR",
+    "not dead"
```
Burnett2k, line 120: 😆

package.json
```diff
@@ -124,6 +113,12 @@
       "@types/eslint": "8.4.3"
     }
   },
+  "browserslist": [
```
Burnett2k, line 116: Optional: It appears this can be placed in a `.browserlistrc` file as well, if we want to keep this config out of package.json (similar to how we have separate .babbelrc.js files.

Burnett2k:
I realize this is still a WIP, but wanted to share that when viewing doc sites I noticed a few issues. There might be more.

1. Live previews don't seem to be working (https://victory-git-babel-preset-env-formidable-labs.vercel.app/gallery/100-column-chart

image

4. Graphics don't seem to be rendering

image

3. the sidebar filter doesn't seem to be working either.

image

KenanYusuf (approved):

Burnett2k (approved):

PR #2807, opened 2024-02-15
Add horizontal stacked bar chart gallery example

### LINKED ISSUE #2798: How would I develop something like this? Is it a VictoryBar?
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Question

I'm looking to develop something similar to this. I know this is accomplished using Victory Charts, but I don't know which component this is. Any direction would be appreciated!

image

carbonrobot:
Add horizontal stacked bar chart gallery example

image

Closes #2798

### DIFF
--- a/docs/src/content/gallery/horizontal-stacked-bars.md
+++ b/docs/src/content/gallery/horizontal-stacked-bars.md
@@ -1,0 +1,59 @@
+---
+id: 51
+title: Horizontal Stacked Bars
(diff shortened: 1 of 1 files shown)


Burnett2k:
Very cool! I saw this question but with my limited knowledge of victory wasn't sure how to accomplish, or what it might be called.

Burnett2k (approved):

PR #2828, opened 2024-03-12
Replace lodash difference with native code

carbonrobot:
Replace lodash difference with native code

### DIFF
--- a/packages/victory-core/src/victory-util/collection.tsx
+++ b/packages/victory-core/src/victory-util/collection.tsx
@@ -34,6 +34,19 @@ export function containsOnlyStrings(
   );
 }
 
+/**
+ * Creates an array of array values not included in the other given arrays
+ * @param a The array to inspect
+ * @param b The values to exclude
+ * @returns The new array of filtered values
+ */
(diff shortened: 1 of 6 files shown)


packages/victory-core/src/victory-util/collection.tsx
```diff
@@ -34,6 +34,19 @@ export function containsOnlyStrings(
...
 
+/**
+ * Creates an array of array values not included in the other given arrays
+ * @param a The array to inspect
+ * @param b The values to exclude
+ * @returns The new array of filtered values
+ */
+export function difference<T>(a: Array<T>, b: Array<T>): Array<T> {
+  if (a && b) {
+    return a.filter((value) => !b.includes(value));
```
Burnett2k, line 45: I thought I could improve performance of this this by using `Set`, but I alas I was wrong. `includes` probably already uses a set internally.

Burnett2k (approved):

packages/victory-core/src/victory-util/collection.tsx
carbonrobot: Not a bad idea!

PR #2829, opened 2024-03-12
Replace lodash identity with native code

carbonrobot:
Replace lodash identity with native code

### DIFF
--- a/packages/victory-core/src/victory-util/axis.tsx
+++ b/packages/victory-core/src/victory-util/axis.tsx
@@ -1,9 +1,18 @@
 import React from "react";
-import { defaults, identity, isObject, uniq, orderBy } from "lodash";
+import { defaults, isObject, uniq, orderBy } from "lodash";
 import * as Collection from "./collection";
 import * as Domain from "./domain";
 import * as Helpers from "./helpers";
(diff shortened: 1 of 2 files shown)


Burnett2k (commented):
Where is this used? Doesn't look like any implementations were updated

packages/victory-core/src/victory-util/axis.tsx
```diff
@@ -1,10 +1,19 @@
...
+import { defaults, isObject, invert, uniq, orderBy } from "lodash";
 import * as Collection from "./collection";
 import * as Domain from "./domain";
 import * as Helpers from "./helpers";
 import { D3Scale } from "../types/prop-types";
 
+/**
+ * Returns the first argument it receives
+ * @param value The value to return
+ * @returns The value give
```
Burnett2k, line 11: nit: should it say `The value given`?

Burnett2k (approved):

packages/victory-core/src/victory-util/axis.tsx
carbonrobot: Fixed.

Burnett2k (approved):

PR #2931, opened 2024-10-28
fix for 'Candlestick labels show incorrect midpoint label'

### LINKED ISSUE #2923: Candlestick labels show incorrect midpoint label
When setting the `labels` property to `true`, candlestick charts incorrectly show the value `true` as a midpoint label, but should show no midpoint label.

Image

<details>
<summary>Example Code</summary>

```
<VictoryChart
  domainPadding={{ x: 25 }}
  theme={VictoryTheme.clean}
>
  <VictoryCandlestick
    labels
    data={[
      {
        x: "3/1/23",
        open: 5,
        close: 10,
        high: 15,
        low: 0,
      },
      {
        x: "3/2/23",
        open: 10,
        close: 15,
        high: 20,
        low: 5,
      },
      {
        x: "3/3/23",
        open: 15,
        close: 20,
        high: 22,
        low: 10,
      },
      {
        x: "3/4/23",
        open: 20,
        close: 10,
        high: 25,
        low: 7,
      },
      {
        x: "3/5/23",
        open: 10,
        close: 8,
        high: 15,
        low: 5,
      },
    ]}
  />
</VictoryChart>
```
</details>

nstolpe:
### Description

Adds a check for a boolean value in label text and does not return a label component if that is the case.

Fixes #2923

### DIFF
--- a/packages/victory-candlestick/src/victory-candlestick.tsx
+++ b/packages/victory-candlestick/src/victory-candlestick.tsx
@@ -28,7 +28,7 @@ export interface VictoryCandlestickStyleInterface {
   data?: VictoryStyleObject;
   high?: VictoryStyleObject;
   highLabels?: VictoryLabelStyleObject | VictoryLabelStyleObject[];
-  labels?: VictoryLabelStyleObject | VictoryLabelStyleObject[];
+  labels?: VictoryLabelStyleObject | VictoryLabelStyleObject[] | boolean;
   low?: VictoryStyleObject;
   lowLabels?: VictoryLabelStyleObject | VictoryLabelStyleObject[];
   open?: VictoryStyleObject;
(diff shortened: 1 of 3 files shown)


packages/victory-candlestick/src/victory-candlestick.tsx
```diff
@@ -28,7 +28,7 @@ export interface VictoryCandlestickStyleInterface {
   data?: VictoryStyleObject;
   high?: VictoryStyleObject;
   highLabels?: VictoryLabelStyleObject | VictoryLabelStyleObject[];
-  labels?: VictoryLabelStyleObject | VictoryLabelStyleObject[];
+  labels?: VictoryLabelStyleObject | VictoryLabelStyleObject[] | boolean;
```
nstolpe, line 31: I don't think this change is necessary, as the label isn't expecting a boolean anywhere else in the code, but leaving it here until review in case there is a desire use case that I'm unfamiliar with.

carbonrobot: Correct, this change isn't needed. This interface is for styling the chart, the `labels` property actually comes from `VictoryMultiLabelableProps`.

carbonrobot (approved):

PR #2934, opened 2024-10-28
fix 'Pie Chart ignores categories prop'

### LINKED ISSUE #2921: Pie Chart ignores categories prop
Pie chart does not correctly order based on the standard categories property.

```jsx
<VictoryPie
  innerRadius={50}
  padAngle={5}
  data={[
    { x: "Cats", y: 30 },
    { x: "Dogs", y: 35 },
    { x: "Birds", y: 25 },
    { x: "Rabbits", y: 10 },
  ]}
  categories={{ x: ["Cats", "Birds", "Dogs", "Rabbits"] }}
  theme={VictoryTheme.clean}
/>
```

image

nstolpe:
### Description

Adds logic to sort pie chart slices by the categories in categories prop. Adds a story as well.

Fixes #2921

### DIFF
--- a/packages/victory-pie/src/helper-methods.ts
+++ b/packages/victory-pie/src/helper-methods.ts
@@ -66,8 +66,10 @@ const sortDataByCategories = (props, data) => {
     const sorted: string[] = [];
     props.categories.x.forEach((category) => {
       const idx = data.findIndex(({ x }) => x === category);
-      const datum = data.splice(idx, 1)[0];
-      sorted.push(datum);
+      if (idx >= 0) {
+        const datum = data.splice(idx, 1)[0];
+        sorted.push(datum);
+      }
     });
     return [...sorted, ...data];
   }
--- a/packages/victory-pie/src/helper-methods.ts
+++ b/packages/victory-pie/src/helper-methods.ts
@@ -59,16 +59,21 @@ const getSlices = (props, data) => {
   return layoutFunction(data);
 };
 
-const sortDataByCategories = (props, data) =>
-  Array.isArray(props?.categories?.x)
-    ? props.categories.x.reduce((newData, category) => {
-        const datum = data.find(({ x }) => x === category);
-        if (datum) {
-          newData.push(datum);
-        }
-        return newData;
-      }, [])
-    : data;
+// sorts data by the categories prop. if all of the data keys aren't included in categories,
+// any remaining data will be appended to the data array.
+const sortDataByCategories = (props, data) => {
+  if (Array.isArray(props?.categories?.x)) {
+    const sorted: string[] = [];
+    props.categories.x.forEach((category) => {
+      const idx = data.findIndex(({ x }) => x === category);
+      const datum = data.splice(idx, 1)[0];
+      sorted.push(datum);
+    });
+    return [...sorted, ...data];
+  }
+
+  return data;
+};
 
 const getCalculatedValues = (props) => {
   const { colorScale, theme } = props;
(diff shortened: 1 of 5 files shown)


packages/victory-pie/src/helper-methods.ts
```diff
@@ -59,6 +59,26 @@ const getSlices = (props, data) => {
...
+      if (idx >= 0) {
+        const datum = data.splice(idx, 1)[0];
+        sorted.push(datum);
+      }
+    });
+    return [...sorted, ...data];
+  }
+
+  return data;
+};
```
carbonrobot, line 80: Looks good... we should also support `categories: string[]` as well, since that is part of the existing spec. Were you able to find where in the code this happens for other charts such as `VictoryBar`? I am assuming we must have some similar code in here.

nstolpe: No mention of categories in `VictoryBar` or it's helper methods, but I'm digging around in some other components to see what I can find. Chart has them and is definitely respecting them, but I'm still looking for the logic that makes it all work.

nstolpe: After digging in to `VictoryChart`, it ends up retrieving and parsing categories through `getCalculatedProps` in its `helper-methods`, which calls `getStringsFromChildren` from the `Wrapper` util, then passes the result of that call ( `{x: string[], y: string[] }`) along with `props` and `childComponents` to `Wrapper.getCategories` to retrieve an object with the `{x: string[], y: string[] }` shape. That object is then passed to `createStringMap` in the helper methods for `VictoryChart`, which returns an object of keys mapped to order (`{ [string]: [number] }`), and then sends that object and other props down to the `VictoryChart`'s child components. It doesn't seem like `PieChart` will have any child components and the code I've added here seems easier to follow to me, but I could use `Wrapper.getCategories` here if you think that makes more sense (I have updated the code in this PR to handle `props.categories` as `string[]`).


I did some more looking into `VictoryBar` too, and it does support categories, I'd been looking for a direct reference in it's code rather than into utility methods. I didn't get as deep into the call chain as with `VictoryChart`, but there are some visual bugs when using scale and categories together, which might be worth opening another issue to address:

```javascript
          <VictoryBar
            theme={VictoryTheme.clean}
            style={{
              data: { fill: VictoryTheme.clean.palette?.colors?.red },
            }}
            scale={{ y: "log", x: "linear" }}
            horizontal
            categories={{
              x: ["a", "b", "c", "d", "e", "f", "g", "h"],
              y: ["i", "j", "k", "l", "m", "n", "o", "p"],
            }}
            data={[
              { x: 1, y: 0.1 },
              { x: 2, y: 1 },
              { x: 3, y: 10 },
              { x: 4, y: 0 },
              { x: 5, y: 0.1 },
              { x: 6, y: 1 },
              { x: 7, y: 10 },
              { x: 8, y: 100 },
            ]}
          />
```
image

Without scale, both x and y categories work fine:

```javascript
          <VictoryBar
            theme={VictoryTheme.clean}
            style={{
              data: { fill: VictoryTheme.clean.palette?.colors?.red },
            }}
            // scale={{ y: "log", x: "linear" }}
            horizontal
            categories={{
              x: ["a", "b", "c", "d", "e", "f", "g", "h"],
              y: ["i", "j", "k", "l", "m", "n", "o", "p"],
            }}
            data={[
              { x: 1, y: 0.1 },
              { x: 2, y: 1 },
              { x: 3, y: 10 },
              { x: 4, y: 0 },
              { x: 5, y: 0.1 },
              { x: 6, y: 1 },
              { x: 7, y: 10 },
              { x: 8, y: 100 },
            ]}
          />
```
image

nstolpe: There's also this build issue I'll need to look into.

carbonrobot: Excellent info, I'll make some notes. The scale issue is expected and I will update the docs.

PR #2942, opened 2024-11-01
Add bundle size analyzer to GH action

nstolpe:
### Description

Adds a bundle-size difference analyzer to Victory so a PRs additions or subtractions from the end bundles size can be observed. Only measures the differences of `packages/victory/dist/victory.min.js`, but can be configured to measure individual package bundles as well. Uses [nejcm/bundle-size-reporter-action](https://github.com/nejcm/bundle-size-reporter-action) for calculations.

#### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### Checklist: (Feel free to delete this section upon completion)

- [ ] I have included a changeset if this change will require a version change to one of the packages.
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have run all builds, tests, and linting and all checks pass
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules


nstolpe:
@carbonrobot this is ready for a look when you have time. It ends up doing 2 builds, one of the branch and one of main, but the whole action completes pretty quickly (possibly thanks to wireit caching, this is my first time working with it). The action hasn't been updated for a while, but it is the most popular in the marketplace so seems to be in wide use.

I've configured it to only measure the `victory.min.js` bundle, but it can measure each minified file if we want/need to go that route. If we did, we'd might want to remove the main bundle from the calculations because it ends up totaling every changed file in the report message.

I was thinking of looking at [bundlemon](https://github.com/LironEr/bundlemon) too. It requires being merged into the parent branch to start reporting, but I think that can be replicated by creating two PRs and parenting the first to the second.

Let me know if you'd like me to investigate bundlemon and/or make any changes here.

nstolpe:
Adding this screenshot before removing the temporary code changes that trigger the comments:
image

carbonrobot (approved):

PR #2948, opened 2024-11-04
Feat: theme builder label options

ramenhog:
### Description

This PR adds the ability adjust the padding, font size and fill color of a theme's axis labels and also includes a refactor to use tailwind instead of sass for theme builder styles. 

I've also removed an unused theme config property for labels since that doesn't seem to actually be used within Victory itself. 

Fixes # (issue)

#### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

### How Has This Been Tested?

Tested on chrome on desktop.


packages/victory-core/src/victory-theme/types.ts
```diff
@@ -194,7 +194,6 @@ export interface VictoryThemeDefinition {
     offsetY?: number;
     orientation?: OrientationTypes;
   } & VictoryCommonThemeProps;
-  label?: VictoryLabelStyleObject | VictoryLabelStyleObject[];
```
ramenhog, line 197: I _believe_ this isn't actually getting read anywhere inside Victory? 

Though, in the future, it might be good to re-evaluate how to refactor the theme config definition so that label styles are set once at the top level of the config and not nested within multiple chart types in the theme config.

carbonrobot: I think we should revert this. I recently filed a bug that this is unused, but _should_ be used in `VictoryLabel`

carbonrobot (approved):

PR #2952, opened 2024-11-05
Split storybook stories into individual files

carbonrobot:
Split storybook stories into individual files for easier maintenance and faster HMR. Improves the development experience by correct Types and Auto gen controls for props.

### Previous format

```
// victory-area.stories.tsx
// cs2 format
export const Basic : () => 
...
export const Theme: () =>
```

### New format
```
// victory-charts/victory-area/basic.stories.tsx
// cs3 format
export const Basic : { ...
```

```
// victory-charts/victory-area/theme.stories.tsx
// cs3 format
export const Theme: { ...
```


nlkluth:
Looks like a few places have `control` set to `"string"`, but I think it should be `"text"`

nlkluth (dismissed):
Playing around a bit with the new file structure and everything seems good. I like it! 🎉

stories/victory-charts/victory-animation/config.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export const ComponentMeta: Meta<StoryProps> = {
+  component: VictoryAnimation,
+  decorators: [componentContainer],
+
+  argTypes: {
+    delay: { control: "number" },
+    duration: { control: "number" },
+    easing: { control: "string" },
```
ramenhog, line 17: Been noticing type errors around `control: "string"`. Looking at the storybook docs, it looks like it should be `control: "text"` instead?

ramenhog (dismissed):
Also liking the new file structure! This makes it much easier to read and maintain. 

Left a comment about a type error I noticed with `control: "string"`. It looks like it's affecting the following files:
- [x] stories/victory-charts/victory-animation/config.ts
- [x] stories/victory-charts/victory-box-plot/config.ts
- [x] stories/victory-charts/victory-candlestick/config.ts
- [ ] stories/victory-charts/victory-chart/config.ts

Also can confirm that the `as const` fixes those arg type errors for me ☝️

ramenhog (approved):
Thanks for fixing! Everything looks generally good now. 

I am still seeing a `themeKey` type issue for certain stories, but I really can't tell why it'd be an issue for one file and not the other. The configs and everything look correct. So it might just be my local 🤷‍♀️

Screenshot 2024-11-07 at 10 37 22 AM

nlkluth (approved):

Issue #2905, opened 2024-10-01
Make bars always visible while zooming


crcarlo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Feature Request

```markdown
If you add a zoom container to a bar chart, bars will disappear as soon as the bar datapoint is outside the new "zoomed" domain.

This makes the the bars appear and disappear as you zoom in and out or pan across the chart.

Can we have the bars rendered if their width is still inside the zoomed domain?
```

carbonrobot:
I created a codesandbox we can use to illustrate the issue. With the default dataset, it seems to work as described. Could you fork the sandbox and add data to reproduce the issue?

https://codesandbox.io/p/sandbox/victory-starter-forked-3zd9hc

crcarlo:
@carbonrobot I think the sanbox is not public as I see a "Sandbox not found" error

crcarlo:
Yes, it's right there! It's easier to see by zooming first and then panning. Once the center of the bar is out of the view the whole bar disappears

https://github.com/user-attachments/assets/a4f28103-efec-4b80-8af0-52fc8f809142

carbonrobot:
Thanks for the video, that describes the issue well. Looks like the clipping path is tied to the grid center point.

PR #2970, opened 2024-11-14
fix/bars-disappear-on-zoom

nstolpe:
### Description

Addresses the issue where bar chart items more than 50% out of their chart area when zoomed disappear completely, as the bar middle point was being used to determine whether they should be culled from view. 

Fixes # [2905](https://github.com/FormidableLabs/victory/issues/2905)

#### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

### Checklist: (Feel free to delete this section upon completion)

- [ ] I have included a changeset if this change will require a version change to one of the packages.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have run all builds, tests, and linting and all checks pass
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

Original video from issue:

https://github.com/user-attachments/assets/311b9725-00b7-4867-a539-90f9dc8e125e

Video after fix:

https://github.com/user-attachments/assets/965e120c-73ee-43bc-9923-33a666e56e85

### DIFF
--- a/packages/victory-bar/src/helper-methods.ts
+++ b/packages/victory-bar/src/helper-methods.ts
@@ -61,6 +61,14 @@ const getCalculatedValues = (props) => {
   let data = Data.getData(props);
   data = Data.formatDataFromDomain(data, domain, 0);
 
+  if (props.groupComponent.type.displayName === 'VictoryClipContainer') {
+    data = data.map((datum) => {
+      datum._x = datum.x;
+      datum._y = datum.y;
+      return datum;
+    });
+  }
+
   return { style, data, scale, domain, origin };
 };
 
--- a/packages/victory-bar/src/helper-methods.ts
+++ b/packages/victory-bar/src/helper-methods.ts
@@ -5,7 +5,7 @@ import {
   Helpers,
   LabelHelpers,
   Scale,
-  VictoryClipContainer
+  VictoryClipContainer,
 } from "victory-core";
 
 export const getBarPosition = (props, datum) => {
--- a/packages/victory-bar/src/helper-methods.ts
+++ b/packages/victory-bar/src/helper-methods.ts
@@ -5,6 +5,7 @@ import {
   Helpers,
   LabelHelpers,
   Scale,
+  VictoryClipContainer
 } from "victory-core";
 
 export const getBarPosition = (props, datum) => {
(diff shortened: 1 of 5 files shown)


packages/victory-bar/src/victory-bar.tsx
```diff
@@ -141,7 +141,12 @@ class VictoryBarBase extends React.Component<VictoryBarProps> {
       return this.animateComponent(props, animationWhitelist);
     }
 
-    const children = this.renderData(props);
+    let children;
+    if (props.groupComponent.type.displayName === 'VictoryClipContainer') {
+      children = this.renderData(props, () => true);
```
nstolpe, line 146: This needs some improvement too. `renderData` has incorrect typing right now (it's type only accepts the `props` parameter), and that always true function can be moved out to a static class property or something else named.

nstolpe:
After seeing @nlkluth's solution [here](https://github.com/FormidableLabs/victory/blob/main/website/docs/charts/box-plot.mdx#box-plot---tooltips), I gave a try at subclassing `VictoryBar` and overriding `renderData`, `getBaseProps` and `getCalculatedValues` from the `EventsMixin` in the derived class to see if this could be accomplished without modifying internal packages (aside from exposing an export or 2). Unfortunately, the overrides got out of control, each additional one triggering an error that required another. I think what's here might be the best way to go at this point.

nlkluth (approved):
These events are tough to parse -- Nice job! One little comment that is not blocking 👍

packages/victory-bar/src/helper-methods.ts
```diff
@@ -61,6 +62,14 @@ const getCalculatedValues = (props) => {
   let data = Data.getData(props);
   data = Data.formatDataFromDomain(data, domain, 0);
 
+  if (props.groupComponent.type === VictoryClipContainer) {
+    data = data.map((datum) => {
```
nlkluth, line 66: Wondering if this should return a new object, of it if doesn't matter since there is already a lot of mutation going on 🤔

nlkluth (dismissed):

packages/victory-bar/src/helper-methods.ts
nstolpe: It definitely won't hurt, I made the update. Thanks.

nlkluth (approved):

PR #3012, opened 2024-12-17
Feat: theme builder export panel updates

ramenhog:
### Description

This PR adds JS and TS outputs for the exported theme. (And fixes a bug where the custom color scale control for stacks were updating the group color scale).

#### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #3007, opened 2024-12-13
Feat: theme builder preview settings

ramenhog:
### Description

This PR includes:
* Consolidating the preview settings into a single dropdown menu for better organization
* Hiding the color scale selector when preview chart doesn't accept an active color scale
* Refactoring the color palette selector into a set of radio buttons. 
  * This allows users to select a palette by clicking on it, which then applies it to the preview. Users can further customize individual colors within the selected palette.

#### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #3004, opened 2024-12-10
Feat: theme builder responsiveness

ramenhog:
### Description

Fixes:
* [Chart grid responsiveness](https://www.notion.so/nearform/Drop-Grid-at-XL-breakpoint-1509aa50dea280c882bfe61a9f9bc38a?pvs=4)
* [Stack config instructions at smaller breakpoints](https://www.notion.so/nearform/Stack-config-instructions-at-smaller-breakpoints-1509aa50dea28071ad10fd8d254840b6?pvs=4)

| | Tablet | Desktop |
| --- | --- | --- |
| Chart Grid | Screenshot 2024-12-10 at 5 00 20 PM | Screenshot 2024-12-10 at 5 00 28 PM |
| Export Panel | Screenshot 2024-12-10 at 5 00 05 PM | Screenshot 2024-12-10 at 4 59 51 PM |

#### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #3017, opened 2024-12-18
feat/theme examples

carbonrobot:
- refactor(theme-builder): move theme builder dir to docs
- refactor(theme-builder): styles to use docs tailwind
- refactor: remove old demo app
- refactor: remove legacy deps from demo app
- refactor: remove demo overrides from eslint
- refactor(theme-builder): use prism-react-renderer instead of react-syntax-highlighter
- refactor: add theme builder to nav and remove demos from linter
- Feat: add theme builder base and box plot chart controls (#2989)
- feat: add color scale overrides and final polish (#2990)
- Feat: theme builder palette fixes (#2992)
- Feat: theme builder new UI (#3000)
- Feat: theme builder responsiveness (#3004)
- Feat: theme builder custom colorscale overrides (#3005)
- Feat: theme builder preview settings (#3007)
- Feat: theme builder refactor axis panel (#3010)
- Feat: theme builder export panel updates (#3012)
- Refactor examples so we can support multiple per chart type


website/src/pages/themes/_providers/previewOptionsProvider.tsx
```diff
@@ -1,19 +1,18 @@
...
   setShowTooltips: (showTooltips: boolean) => void;
   resetPreviewOptions: () => void;
-  exampleConfigs: ControlConfig[];
-  setExampleConfigs: (exampleConfigs: any[]) => void;
+  exampleContent: ExampleConfig[];
+  setExampleContent: (content: ExampleConfig[]) => void;
 };
 
-export const defaultColorScale = undefined;
+export const defaultColorScale = "qualitative";
```
ramenhog, line 14: I believe this should still be undefined to handle the case where the `colorScale` prop is not explicitly set on the Victory component. When that happens, the Victory component will use the custom color scale that's set for that chart type in the theme config (i.e. `theme.pie.colorScale`)

website/src/pages/themes/_components/sidenav.tsx
```diff
@@ -49,48 +50,54 @@ export const NAV_ITEMS: NavItem[] = [
     title: "Base Theme",
     icon: CircleStackIcon,
     panelType: "theme",
+    content: AllExamples,
```
ramenhog, line 53: Nice! 💯

ramenhog:
@carbonrobot so I played around with a few layouts for the color palette panel and this is what I ended up with:



It's still slightly unconventional UX 😅 but at least this way, the preview is always showing the user what the user is currently customizing and there's ways to go back and forth between seeing defaults and selected color scales. 

Whatcha think? If you still prefer a dropdown up top and no radio buttons, we can do that. But when I had that layout, the number of controls felt more overwhelming and the user would have to go back up to adjust their preview color scale:

PR #3031, opened 2025-01-10
feat(theme-builder): add warning message when leaving theme builder

ramenhog:
### Description

This PR adds a warning alert when users try to navigate away from theme builder.
https://www.notion.so/nearform/Display-prompt-when-navigating-away-mid-edit-1759aa50dea280358b9ef7f65bb45dcb?pvs=4

### DIFF
--- a/website/src/pages/themes/index.tsx
+++ b/website/src/pages/themes/index.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 
 import "../../css/custom.css";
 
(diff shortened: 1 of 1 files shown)


nlkluth (approved):

carbonrobot (changes requested):
I'm not sure I am aligned on this one.

Generally, I think we should not use a `window.confirm` since users are conditioned to click "OK", which loses their changes when that may not be the intent. In some browsers it also blocks all browser interaction, including closing the tab or window (even during a system OS restart).

But more importantly, this means that for any user, regardless of whether they didn't make changes, or if they did and already exported them, they will see this confirmation box.

Perhaps an alternative would be store/sync  changes (and only changes) in a local state or local session using something like `useLocalStorage` hook instead of the `useState` where we store the current data?

ramenhog:
@carbonrobot valid point! I like the local storage option as an alternative solution so I'll work on that approach instead 👍

